PR #10, opened 2019-10-28
feat(lean-depot): deployment of curated sets of Lean libraries

cipher1024:
# Curated libraries

A list of library versions is compiled, checked and published at https://gitlab.com/simon.hudon/lean-depot/blob/snapshot/3.4.2/snapshot.toml

Library authors can submit their library by creating a configuration file for it and create a pull request at https://gitlab.com/simon.hudon/lean-depot/tree/master/pkgs

# `lean-depot.toml` configuration files

`lean-depot.toml` configuration files list libraries without specifying a version for each. They also specify a snapshot of the Lean ecosystem. That snapshot selects versions of packages that have been checked to work together.

From a `lean-depot.toml` file, a `leanpkg.toml` is generated and the normal Lean workflow can proceed from there.

# Using lean-depot

`lean-depot` can be installed using the normal mathlib-tools setup.

An existing `leanpkg` project can be complemented with a `lean-depot.toml` configuration file by calling `lean-depot init`. The dependencies will be copied from the `leanpkg.toml` file.

A new `lean-depot` project by calling `lean-depot new NAME`. It will create a `lean-depot.toml` file and a `leanpkg.toml` file and configure the dependencies.

`lean-depot list` will enumerate the available libraries and `lean-depot add MY_LIB` will add `MY_LIB` to the dependencies.


semorrison:
Could you just explain first what problem you’re trying to solve, that the
existing leanpkg.toml files don’t address? I’m confused why we’re adding
another package management file in the first place.

scripts/lean-depot.py
```diff
@@ -0,0 +1,293 @@
+#! /usr/local/bin/python3
```
rwbarton: This isn't a standard location.

scripts/lean-depot.py
```diff
@@ -0,0 +1,293 @@
...
+    res = toml.decoder.TomlDecoder()
+    res = res.get_empty_inline_table()
+    res.update(d)
+    return res
+
+def is_prefix_of(x,y):
+    return x == y[:len(x)]
+
+def snapshot_name(x):
+    (base,ext) = os.path.splitext(x)
```
rwbarton: Presumably `x`, whatever that is, is expected to conform to some specific format. Is that documented anywhere?

scripts/lean-depot.py
```diff
@@ -0,0 +1,293 @@
...
+    if base == 'snapshot':
+        return 'nightly'
+    else:
+        return base[len('snapshot')+1:]
+
+def get_depot(vers):
```
rwbarton: Longer variable names would help, e.g., `vers` -> `lean_version` (I think).

scripts/lean-depot.py
```diff
@@ -0,0 +1,293 @@
...
+    depot = [ (k, load_snapshot(v))
+                  for (k,v) in depot.items() ]
+                          # key=lambda x: x[0])
+    depot = [ k
+              for (k,v) in depot
+              if len([ k2 for k2 in deps.values()
+                       if k2 not in v ]) == 0 ]
+    (x,depot) = (depot[0], depot[1:])
+    if len(depot) == 0: snap = x
+    else: snap = max(depot)
```
rwbarton: I don't follow the logic here.

scripts/lean-depot.py
```diff
@@ -0,0 +1,293 @@
...
+    """setup leanpkg configuration"""
+
+    prj = toml.load('lean-depot.toml')
+    tag  = prj['package']['lib_snapshot']
+    vers = prj['package']['lean_version']
+    depot = toml.load(get_depot(vers)[tag])
+    # print(depot)
+    # print(prj)
+    name         = prj['package']['name']
+    lean_version = prj['package']['lean_version']
```
rwbarton: Did it change since a few lines above?

scripts/lean-depot.py
```diff
@@ -0,0 +1,293 @@
...
+        h.write('#   * edit lean-depot.toml and not this file\n')
+    os.system('leanpkg configure')
+
+@user_command('init')
+def init():
+    """create lean-depot configuration from leanpkg file"""
+    leanpkg = toml.load(f'leanpkg.toml')
+    name         = leanpkg['package']['name']
+    lean_version = leanpkg['package']['lean_version']
+    lean_version = str.split(lean_version,':')[-1]
```
rwbarton: Do Lean versions tend to contain `:`? If this is needed here then why not elsewhere as well?

scripts/lean-depot.py
```diff
@@ -0,0 +1,293 @@
...
+    depot = get_depot(lean_version)
+    if len(depot) <= 1:
+        snapshot = 'nightly'
+    else:
+        snapshot = max(filter(lambda x: x != 'nightly', depot.keys()))
+    all_deps = leanpkg['dependencies']
+    deps = { k : v['git'][0] for (k,v) in all_deps.items()
```
rwbarton: I don't understand this `[0]`. Isn't `v['git']` going to be something like `'https://github.com/leanprover/mathlib'`, so that `v['git'][0]` is `h`?

scripts/lean-depot.py
```diff
@@ -0,0 +1,293 @@
...
+             for fn in os.listdir('pkgs')
+             if os.path.splitext(fn)[1] == '.toml' }
+    print(pkgs)
+
+    snap = { 'snapshot' : { k : { 'git' : pkg['url'],
+                                  'rev' : pkg['commit'],
+                                  'desc' : pkg['description'] }
+                                for (k,pkg) in pkgs.items() } }
+    with open(path, 'w') as h:
+        print(toml.dump(snap, h))
```
rwbarton: `print(toml.dump(...))` is odd.

scripts/lean-depot.py
```diff
@@ -0,0 +1,293 @@
...
+             f(**args.__dict__)
+         elif len(argv) == 2:
+             f()
+         else:
+             usage()
+    else:
+        usage()
+
+# Local Variables:
+# eval: (add-hook 'after-save-hook (lambda () (copy-file (buffer-file-name) "~/.mathlib/bin/lean-depot" t)) nil t)
```
rwbarton: I think this will be annoying to other people editing the file. Why not use a symlink?

scripts/remote-install-update-mathlib.sh
```diff
@@ -57,24 +57,26 @@ pip3 install --upgrade $USER setuptools || exit -1
 pip3 install --upgrade $USER $PYTHON_DEPS || exit -1
 echo "Fetching the update-mathlib script"
 curl -o update-mathlib.py https://raw.githubusercontent.com/leanprover-community/mathlib-tools/$BRANCH/scripts/update-mathlib.py
+curl -o update-mathlib.py https://raw.githubusercontent.com/leanprover-community/mathlib-tools/$BRANCH/scripts/lean-depot.py
```
rwbarton: Output filename is wrong.

scripts/remote-install-update-mathlib.sh
```diff
@@ -57,24 +57,26 @@ pip3 install --upgrade $USER setuptools || exit -1
...
 curl -o auth_github.py https://raw.githubusercontent.com/leanprover-community/mathlib-tools/$BRANCH/scripts/auth_github.py
 curl -o post-commit https://raw.githubusercontent.com/leanprover-community/mathlib-tools/$BRANCH/scripts/post-commit
 curl -o post-checkout https://raw.githubusercontent.com/leanprover-community/mathlib-tools/$BRANCH/scripts/post-checkout
 echo "installing it in \$HOME/.mathlib/bin"
-chmod +x update-mathlib.py
-mkdir -p "$HOME"/.mathlib/bin || true
-mkdir -p "$HOME"/.mathlib/hooks || true
+chmod +x update-mathlib.py cache-olean.py lean-depot.py
+mkdir -p $HOME/.mathlib/bin || true
+mkdir -p $HOME/.mathlib/hooks || true
```
rwbarton: I don't know why the double quotes went away here and below, but they are definitely needed.

scripts/setup-dev-scripts.sh
```diff
@@ -78,18 +78,20 @@ then
...
-	cp delayed_interrupt.py    "$HOME"/.mathlib/bin/
-	cp update-mathlib.py       "$HOME"/.mathlib/bin/update-mathlib
-	cp cache-olean.py          "$HOME"/.mathlib/bin/cache-olean
-	cp setup-lean-git-hooks.sh "$HOME"/.mathlib/bin/setup-lean-git-hooks
-        chmod +x "$HOME"/.mathlib/bin/update-mathlib
-        chmod +x "$HOME"/.mathlib/bin/cache-olean
-        chmod +x "$HOME"/.mathlib/bin/setup-lean-git-hooks
-	cp post-commit   "$HOME"/.mathlib/hooks/
-	cp post-checkout "$HOME"/.mathlib/hooks/
+	mkdir -p $HOME/.mathlib/bin || true
```
rwbarton: Again the old double quotes were correct.

scripts/lean-depot.py
cipher1024: `snapshot-2019-10.toml` for the October snapshot

rwbarton:
Scott, I can try to explain, being somewhat familiar with analogous systems for other languages (e.g. [Stackage](https://www.stackage.org/)).

Suppose you've just proved Scott's Amazing Theorem and you want to formalize it in Lean. As it happens, it only requires the definition of a perfectoid space, some facts about weak homotopy equivalences, and the independence of the continuum hypothesis. How fortunate!

The first problem you would have run into (at least until recently) is that there probably are no versions of the libraries you want to use that all build against the same version of mathlib. Or even if there are such versions, it is difficult to find them. That is solved by designating certain mathlib versions as "releases" or snapshots and encouraging authors of other packages to release versions which build against the mathlib snapshots. For example the three referenced packages all now have versions which build against a particular version of mathlib from early October.

Now you will want to specify in your `leanpkg.toml` file that you want to build your project against those specific versions of each of the three packages. You could manually go look what versions they are and specify their hashes in `leanpkg.toml`, and it would probably only take a few minutes. But it's still annoying, doesn't really clarify your intent and makes it hard for someone else (like later you) to figure out what the right thing to do is if they want to add a fourth package that is part of the same snapshot. Writing the snapshot version and library names to `lean-depot.toml` expresses the intent and eliminates the drudgery.

For now it might seem like overkill because we have few packages (and my hypothetical scenario is rather farfetched), but in the future, especially if Lean is used more as a programming language, it will be more common to have a package with a lot of other dependencies and want to not have to manage them by hash manually.

scripts/lean-depot.py
cipher1024: Thanks for the suggestion

cipher1024:
Thanks for the nice summary. I would add that the small number of Lean packages might also be a chicken and egg problem. Discussing the issue with other Lean users, I've been repeatedly answered that the best solution is to put everything in mathlib. But that is a very centralized and heavy weight solution. If you write your own package, you can adopt your own standards and design principles.

semorrison:
Thanks. If I understand correctly, if we regained control of the
leanpkg.toml format this could all be done inside that.

cipher1024:
I think an argument could be made for keeping `lean` and the package management separate. You may have to adopt new features for the package even when using old `lean` versions.

cipher1024:
@rwbarton You may notice that, compared to the last time I talked to you about it, I have changed the way one submits a library to `lean-depot`. Instead of creating a tag in your own repo, you submit a PR with a configuration file for your library. This allows me to implement the `auto_update` feature where, upon request, `lean-depot` will routinely try the last commit (or maybe I should make it the head of a certain branch) of your library when creating a new snapshot. I'm hoping that this will reduce the burden for library maintainers to keep their package working with the rest of `lean-depot`.

cipher1024:
And now that I think of it, `cache-olean` and `update-mathlib` could be generalized and integrated into `lean-depot`

cipher1024:
@rwbarton Did I address your concerns? Also, I'm thinking of emailing package authors automatically when their package doesn't build. The alternative might be to automatically create an issue on the github repo of the package. Do you have any thoughts on that?

jcommelin:
Sounds good. And you can make it a config option:
```
  contact-on-build-failure: email, // alternative: github
```

cipher1024:
Great idea @jcommelin! Feel free to PR this information for the perfectoid project if you want to be contacted by email along with the `author` field. Otherwise, we should default to using github.

It's not implemented yet but having the information there won't hurt :)

jcommelin:
@cipher1024 I would prefer getting contacted via github. So then I don't do anything, right?

cipher1024:
That's right

rwbarton:
> @rwbarton Did I address your concerns? Also, I'm thinking of emailing package authors automatically when their package doesn't build. The alternative might be to automatically create an issue on the github repo of the package. Do you have any thoughts on that?

To be honest, I probably wouldn't have commented at all if I hadn't seen that a bunch of double quotes in shell scripts were incorrectly removed. But please don't use this as a tactic in the future :)

The main thing needed here is system documentation.
* What is a snapshot?
* What is the format of the file defining a snapshot?
* Where do those files live?
* What is the syntax of a snapshot identifier?
* What does `nightly` mean?
* What is the syntax of a `lean-depot.toml` file?
* How does a package author handle packages which are not available from lean-depot?
* What is the workflow for managing `lean-depot.toml` and `leanpkg.toml` files? What should be checked into VC?
* What is the process for maintainers to have their packages added to lean-depot?

I was able to infer answers to some of these questions by reading `lean-depot.py` itself, but that script should not be the "source of truth" for these answers.

PatrickMassot:
I'm closing this since it's been inactive for almost one year, and Rob implemented another variation on this idea, but feel free to reopen if you want to work on it again.

PR #11, opened 2019-12-21
Add safety check for dirty repo

ocfnash:
TO CONTRIBUTORS:

Make sure you have:

  * [ ] reviewed and applied the coding style: [coding](https://github.com/leanprover/mathlib/blob/master/docs/style.md), [naming](https://github.com/leanprover/mathlib/blob/master/docs/naming.md)
  * [ ] for tactics:
     * [ ] added or adapted documentation in [tactics.md](https://github.com/leanprover/mathlib/blob/master/docs/tactics.md)
     * [ ] write an example of use of the new feature in [tactics.lean](https://github.com/leanprover/mathlib/blob/master/tests/tactics.lean)
  * [ ] make sure definitions and lemmas are put in the right files
  * [ ] make sure definitions and lemmas are not redundant

If this PR is related to a discussion on Zulip, please include a link in the discussion.

For reviewers: [code review check list](https://github.com/leanprover/mathlib/blob/master/docs/code-review.md)

### DIFF
--- a/scripts/cache-olean.py
+++ b/scripts/cache-olean.py
@@ -75,6 +75,15 @@ def fetch_mathlib(asset):
         ar.extractall('.')
         print("... successfully extracted olean archive.")
 
+def should_proceed_with_dirty_repo(action):
+    if action == ['--fetch-even-if-dirty']:
+        return True
+    if action == ['--fetch']:
+        print('Your repo is dirty; fetching in this state could cause you to lose data. '\
+              'If you understand the risks, you can use --fetch-even-if-dirty.')
+    else:
+        print('Your repo is dirty; you must commit or discard your changes to perform this action.')
+    return False
 
 if __name__ == "__main__":
     try:
@@ -88,8 +97,7 @@ if __name__ == "__main__":
         sys.exit(-1)
 
     if repo.is_dirty():
-        if sys.argv[2:] != ['--even-if-dirty'] and\
-        input('Your repo is dirty. You may lose data if you proceed. Proceed? [N] ').lower() != 'y':
+        if not should_proceed_with_dirty_repo(sys.argv[1:]):
             sys.exit(-1)
         print('Warning: proceeding with dirty repo.')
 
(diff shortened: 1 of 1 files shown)


ocfnash:
cf this thread (separated across two URLs due to rename): [part 1](https://leanprover-community.github.io/archive/113489newmembers/08888cacheoleandangerous.html) [part 2](https://leanprover-community.github.io/archive/113489newmembers/49795cacheoleanuserconfusion.html)

scripts/cache-olean.py
```diff
@@ -87,6 +87,12 @@ def fetch_mathlib(asset):
         print('Repository not initialized')
         sys.exit(-1)
 
+    if repo.is_dirty():
```
cipher1024: I think the flag should only work with `--fetch`. And with fetch when it is absent and the directory is dirty, I would print an error message and exit with code -1

With `build`, `build-all` and without flag (i.e. `cache`) if the directory is dirty, I would print an error and suggest the user commits the changes before building without possibility of bypassing the behavior. Because the cache is indexed with the hash code of the commit, if the directory is dirty, we're storing the wrong files.

ocfnash: Thanks for the feedback, I agree this would be better. I have just added a commit which attempts to implement the scheme you have sketched. :-)

scripts/cache-olean.py
```diff
@@ -75,6 +75,15 @@ def fetch_mathlib(asset):
         ar.extractall('.')
         print("... successfully extracted olean archive.")
 
+def should_proceed_with_dirty_repo(action):
+    if action == ['--fetch-even-if-dirty']:
+        return True
+    if action == ['--fetch']:
+        print('Your repo is dirty; fetching in this state could cause you to lose data. '\
+              'If you understand the risks, you can use --fetch-even-if-dirty.')
```
cipher1024: I'd change the second line of the message to `'To proceed regardless, use --fetch-even-if-dirty.'`

scripts/cache-olean.py
```diff
@@ -75,6 +75,15 @@ def fetch_mathlib(asset):
...
+    else:
+        print('Your repo is dirty; you must commit or discard your changes to perform this action.')
```
cipher1024: Use `'Your repo is dirty; commit or discard your changes to perform this action.'` instead.

cipher1024:
Because of github access rights, the build was failing so I copied the branch over to this repo to test it properly

PR #18, opened 2020-03-06
fix(mathlib/leanproject): fix leanproject get command

gebner:
`leanproject get` used to fail because the argument `target` was different than the `click.argument` `dir`.


PatrickMassot:
Thanks, but this is no longer necessary.

PR #23, opened 2020-03-06
Refactor reloaded

PatrickMassot:
This is PR #16 coming back after reverting a premature merge.


gebner:
@PatrickMassot I've updated the `default.nix` file.

PatrickMassot:
Thanks Gabriel. Actually this is what I meant to do in my previous default.nix commit, but I was doing too many things at once. I rewrote history.

sgouezel:
One more feature request (no urgency, this can be postponed): I hit Ctrl-C during the olean downloading in a run of `leanproject up`. Then, if I do `leanproject up`, it realizes that the archive is corrupt, but it does not try to download it again:
```
> leanproject up
Looking for local mathlib oleans
Found local mathlib oleans
Compressed file ended before the end-of-stream marker was reached
```
Is is possible to download the archive again in this situation?

jcommelin:
@sgouezel What happens if you force downloading the cache using `--force-download`?

sgouezel:
I can definitely do it by hand, for instance by removing the downloaded file (or possibly using `--force-download`, I haven't tried). But I think that for the user who doesn't know what is going on, automatic recovery would be handy.

PatrickMassot:
@sgouezel Fixed in 8cb119c.

mathlibtools/lib.py
```diff
@@ -357,9 +357,16 @@ def get_mathlib_olean(self) -> None:
         if not (self.directory/'leanpkg.path').exists():
             self.run(['leanpkg', 'configure'])
         self.mathlib_folder.mkdir(parents=True, exist_ok=True)
-        unpack_archive(get_mathlib_archive(self.mathlib_rev, self.cache_url,
+        try:
+            unpack_archive(get_mathlib_archive(self.mathlib_rev, self.cache_url,
                                            self.force_download), 
                        self.mathlib_folder)
+        except (EOFError, shutil.ReadError):
+            log.info('Something wrong happenend with the olean archive. '
```
bryangingechen:
```suggestion
            log.info('Something wrong happened with the olean archive. '
```

PR #32, opened 2020-03-24
Don't bump GitPython version all that much

Kha:
2.1.11 is the oldest version containing https://github.com/gitpython-developers/GitPython/commit/7f08b7730438bde34ae55bc3793fa524047bb804, and coincidentally the version available in Nixpkgs stable.


Kha:
Usually I prefer to pin the Nixpkgs commit to avoid such issues (unstable has 3.0.5), but this seemed like a no-brainer fix.

PR #35, opened 2020-03-28
Auth

PatrickMassot:
This restores trying to use Github authentication when accessing legacy olean caches. This should also get rid of the warning about not being in a Git repository.

I find it a bit annoying to open a PR for this because master is protected. Who has the power to change that?


robertylewis:
You now have admin privileges on this repo, so you can change whatever you want.

PR #38, opened 2020-03-29
Add command import-graph

PatrickMassot:
This is experimenting import graph generation in `leanproject`. The only downside I can see is adding even more dependencies to mathlib tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,8 @@ structure, you can run:
 leanproject import-graph my_graph_file_name.suffix
 ```
 where the suffix will determine the output format. It must be either
-`dot` or `graphml` or `gexf`.
+`dot` or `graphml` or `gexf`, (or `pdf`, `svg` or `png` if
+[graphviz](https://www.graphviz.org/) is installed).
 If you want to restrict the graph to files leading to a certain file
 `my_subproject/my_file.lean` then you can run:
 ```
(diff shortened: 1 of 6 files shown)


semorrison:
I pushed some documentation improvements. I'll investigate adding the functionality I want next.

semorrison:
Okay, I've completely failed to use [bfs_tree](https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.algorithms.traversal.breadth_first_search.bfs_tree.html), mostly because I don't understand python well enough, so I'm quitting on this for now.

I think it's great that we now have a tool to generate import graphs, although I don't think this usefully solves my original use case "why on earth is changing X causing Y to compile?". I could use the `--to` and `--from` arguments, but without a text mode answer it's too much overhead to fire up a graph visualisation program, just to inspect a linear chain of identifiers.

semorrison:
Also, if anyone wants to tweak this further:

* I think `--to` and `--from` are still backwards. The example Patrick gave on zulip was
```
leanproject import-graph --to analysis.normed_space.hahn_banach --from category_theory.category.default hb.pdf
```
which to me read as finding the ways in which `category.default` depends on the Hahn-Banach theorem (which would be even more unusual!).

* For understanding "why on earth is changing X causing Y to compile?", it would be better to intersect the ancestors and descendants graphs, rather than using `path`. If you're considering tree to prune the dependency tree, it's essential to know ahead of time how many cuts you'd need to make!

bryangingechen:
@semorrison Did you mean to fold the build-time-bot scripts into this PR?

PR #39, opened 2020-04-02
Raise exceptions on leanpkg and git errors

Vtec234:
When running `leanproject` in CI (and arguably in general), we want it to exit with a non-zero code if the command fails. Here, I make calls to `subprocess.run` raise an exception if the tool fails where that makes sense, so that the exception handler is called and `leanproject` overall exits with an error. Otherwise, builds that clearly failed ([e.g. here](https://github.com/b-mehta/topos/runs/556155330)) still get a green tick.

### DIFF
--- a/mathlibtools/leanproject.py
+++ b/mathlibtools/leanproject.py
@@ -163,8 +163,7 @@ def get_project(name: str, directory: str = '') -> None:
         name = name + '_' + branch
     directory = directory or name
     if directory and Path(directory).exists():
-        log.error(directory + ' already exists')
-        sys.exit(-1)
+        raise FileExistsError('Directory ' + directory + ' already exists')
     try:
         LeanProject.from_git_url(url, directory, branch,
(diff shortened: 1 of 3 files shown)


PatrickMassot:
I'm sorry I've neglected this PR, I should have merged it before the import graph stuff. Could you rebase and force push? I'll merge it right away.

Vtec234:
@PatrickMassot done!

PatrickMassot:
GitHub still complains "This branch cannot be rebased due to conflicts".

Vtec234:
That's strange, there's a green tick here. I also just tried `git merge errors` into a copy of current `master` locally and that works. Are you trying to merge this PR via the squash or the merge method? Perhaps it's just the rebase way that doesn't work.

PatrickMassot:
Indeed I didn't notice I could switch from "rebase and merge" to "squash and merge". Sorry about the noise.

PR #43, opened 2020-04-14
Get commands executable path

mateoatr:
Calling commands such as `leanpkg` through `subprocess.run([cmd])` will not work on Windows, since this will not search for the given command in the `PATH`. The PR fixes this issue.

/cc @PatrickMassot

### DIFF
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -79,7 +79,7 @@ VersionTuple = Tuple[int, int, int]
 
 def mathlib_lean_version() -> VersionTuple:
     """Return the latest Lean release supported by mathlib"""
-    out = subprocess.run(['git', 'ls-remote', '--heads', MATHLIB_URL],
+    out = subprocess.run([shutil.which('git'), 'ls-remote', '--heads', MATHLIB_URL],
             stdout=subprocess.PIPE, check=True).stdout.decode()
     version = (3, 4, 1)
     for branch in out.split('\n'):
@@ -233,17 +233,17 @@ class ImportGraph(nx.DiGraph):
         """Writes itself to a graphviz dot file."""
         path = path or self.base_path/'import_graph.dot'
         nx.drawing.nx_pydot.to_pydot(self).write_dot(str(path))
-    
+
     def to_gexf(self, path: Optional[Path] = None) -> None:
         """Writes itself to a gexf dot file, suitable for Gephi."""
         path = path or self.base_path/'import_graph.gexf'
         nx.write_gexf(self, str(path))
-    
+
     def to_graphml(self, path: Optional[Path] = None) -> None:
         """Writes itself to a gexf dot file, suitable for yEd."""
         path = path or self.base_path/'import_graph.graphml'
         nx.write_graphml(self, str(path))
-    
+
     def write(self, path: Path):
         if path.suffix == '.dot':
             self.to_dot(path)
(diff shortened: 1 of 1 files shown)


mathlibtools/lib.py
```diff
@@ -522,16 +522,20 @@ def run(self, args: List[str]) -> str:
...
                               stdout=subprocess.PIPE,
                               check=True).stdout.decode()
-    
+
     def run_echo(self, args: List[str]) -> None:
         """Run a command in the project directory, letting stdin and stdout
         flow.
 
            args is a list as in subprocess.run"""
+        if len(args):
```
PatrickMassot, line 537: This test is useless (and it should simply be `if args:` if it were useful) because `subprocess.run` would raise an exception anyway.

mathlibtools/lib.py
```diff
@@ -522,16 +522,20 @@ def run(self, args: List[str]) -> str:
         """Run a command in the project directory, and returns stdout + stderr.
 
            args is a list as in subprocess.run"""
-        return subprocess.run(args, cwd=str(self.directory), 
+        if len(args):
```
PatrickMassot, line 525: This test is useless (and it should simply be if args: if it were useful) because subprocess.run would raise an exception anyway.

mateoatr:
I tested this locally and it solved the problem, which is why I put the PR... I already reproed the issue in two machines using win10, python 2.7 and 3.8, and pip versions > 19.

mateoatr:
The hassle of introducing this changes just for the sake of fixing 'half-buggy environments' might well not be worth it, especially when there's options such as WSL to avoid any cmd shenanigans. Feel free to discard, thanks.

PatrickMassot:
I think I haven't been clear enough. I don't mind the "hassle of introducing this changes", especially since you opened a PR instead of complaining without helping. What was wrong was boldly claiming something "will not work on Windows" despite the project being used by many people on Windows, and having continuous integration testing every new commit on Windows.

Now there is another issue with this PR. It's partly my fault because the contributing guide was hidden at the bottom of the README. I've now moved it, with added explanations, at the [standard location](https://github.com/leanprover-community/mathlib-tools/blob/master/CONTRIBUTING.md).

Your code fails static type checking:
```
$ mypy mathlibtools
mathlibtools/lib.py:82: error: List item 0 has incompatible type "Optional[str]"; expected "Union[bytes, str, _PathLike[Any]]"
mathlibtools/lib.py:260: error: List item 0 has incompatible type "Optional[str]"; expected "Union[bytes, str, _PathLike[Any]]"
mathlibtools/lib.py:438: error: List item 0 has incompatible type "Optional[str]"; expected "str"
mathlibtools/lib.py:498: error: List item 0 has incompatible type "Optional[str]"; expected "str"
mathlibtools/lib.py:508: error: List item 0 has incompatible type "Optional[str]"; expected "Union[bytes, str, _PathLike[Any]]"
mathlibtools/lib.py:512: error: List item 0 has incompatible type "Optional[str]"; expected "Union[bytes, str, _PathLike[Any]]"
mathlibtools/lib.py:525: error: Incompatible types in assignment (expression has type "Optional[str]", target has type "str")
mathlibtools/lib.py:536: error: Incompatible types in assignment (expression has type "Optional[str]", target has type "str")
mathlibtools/lib.py:563: error: List item 0 has incompatible type "Optional[str]"; expected "str"
mathlibtools/lib.py:594: error: List item 0 has incompatible type "Optional[str]"; expected "str"
mathlibtools/lib.py:607: error: List item 0 has incompatible type "Optional[str]"; expected "str"
mathlibtools/lib.py:647: error: List item 0 has incompatible type "Optional[str]"; expected "str"
Found 12 errors in 1 file (checked 5 source files)
```
The easy solution is to add near the top of `lib.py`:
```python
def which(cmd: str) -> str:
    """Find the path to cmd using which or raise ValueError"""
    path = shutil.which(cmd)
    if path:
        return path
    else:
        raise ValueError('Command {} not found'.format(cmd))
```
and use that instead of `shutil.which` everywhere else.

PatrickMassot:
I'm closing this since it seems abandoned by the author and no longer cleanly merge. Feel free to reopen after resolving conflicts and fixing typing issues as I explained in my previous comment.

PR #61, opened 2020-06-28
Add private SSH key handling in leanproject get

pechersky:


pechersky:
From discussion in Zulip: https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/leanproject.20get.20mathlib.2C.20https.20vs.20ssh

PR #63, opened 2020-06-30
Use pipx for installation of mathlibtools

TwoFX:
This is a companion to leanprover-community/leanprover-community.github.io#77, synchronizing the detailed instructions with the automatic script.


PatrickMassot:
@TwoFX I stupidly assumed you had tested this change. I'm now going to revert it.  
FYI, trying to use the install script currently gives:

image

then launching a new terminal and trying to install mathlibtools gives:
image
This is all on a fresh Ubuntu 20.04 install.

Thanks again for trying to help. Feel free to reopen a PR once you have a procedure that actually works.

PR #65, opened 2020-07-03
feat(options): hide aliases for upgrade-mathlib, and allow prefixes

semorrison:
This PR:

1. hides the aliased commands `up` and `update-mathlib`, only showing `upgrade-mathlib` in the help text
2. allows users to use unique prefixes of commands

In particular, this means that now any of `leanproject up`, `leanproject update`, `leanproject upgrade`, `leanproject update-mathlib`, `leanproject upgrade-mathlib`, will all do as expected.

`leanproject b` will run `leanproject build`, but `leanproject d` will give an error message:

```
Try 'leanproject -h' for help.

Error: Too many matches: decls, delete-zombies
```

### DIFF
--- a/mathlibtools/leanproject.py
+++ b/mathlibtools/leanproject.py
@@ -35,6 +35,7 @@ class CustomMultiCommand(click.Group):
                     cmd = super(CustomMultiCommand, self).command(
                         alias, *args[1:], **kwargs)(f)
                     cmd.short_help = "Alias for '{}'".format(_args[0])
+                    cmd.hidden = True
             else:
                 _args = args
             cmd = super(CustomMultiCommand, self).command(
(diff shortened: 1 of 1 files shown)


semorrison:
(This was prompted by my incorrectly advising someone on zulip to try `leanproject upgrade`.)

PatrickMassot:
Thanks Scott. I guess the argument for not doing that is every time we'll add a new command users may need to reconfigure their muscle memory if a prefix suddenly becomes ambiguous. But I don't have a strong opinion here. Let's wait a bit to see what others think. If this is merged then don't hesitate to remind me to release.

If someone has time and energy for this kind of thing, there is one issue that I've been meaning to address since day one, but never got motivated enough: we should investigate how to setup shell completion for leanproject commands. The painful part is to figure out what is a reasonable set of shells to support and whether this can be part of the installation process or a new command `leanproject install-completions` or whatever.

PR #66, opened 2020-07-08
Use pipx for installation of mathlibtools

TwoFX:
See #63. I have verified that this script works on fresh installs of Debian 10.3 and Ubuntu 20.04.


PatrickMassot:
Thanks! I can confirm it works on my fresh Ubuntu 20.04 VM.

Issue #72, opened 2020-07-23
SSH problems


mattearnshaw:
the block of code at https://github.com/leanprover-community/mathlib-tools/blob/77c707eaead3732d0a1527478602649c7e82874e/mathlibtools/leanproject.py#L180 seems to exist to test whether an SSH URL (`git@...`) should be used to clone a project when the latter is given by name. But this test breaks the tool because my `ssh_config` is set to use a specific identity file for the `github.com` host, which `paramiko` does not seem to find automatically

I think perhaps `https` URLs should just be used as a default if a protocol is not specified, otherwise here is an alternative that does an `ssh_config` lookup but the user is potentially prompted twice for their key (once by `paramiko`, then again by `git`): https://github.com/mattearnshaw/mathlib-tools/commit/a6193c7a7144ef8765d9e2b56d633ddf29e1f4bc

gebner:
> I think perhaps `https` URLs should just be used as a default if a protocol is not specified

This is unfortunately not an option because you can't push via `https` if you have 2FA enabled.  (And I am required to enable it because of another project.)

I think parsing the `ssh_config` is the right way to go here.  We also use the public key for `git clone` etc., do we need to change anything there as well to support host-specific keys?

mattearnshaw:
a change is only needed where `paramiko` is used so I think my branch covers it – `git` (or the python module for git) handles the host-specific keys properly by default. By using the `fabric` library, we also don't need to add our own `ssh_config` parsing: https://github.com/leanprover-community/mathlib-tools/compare/master...mattearnshaw:ssh_keyfile

Another option would be to include a flag for `ssh` such that e.g. `leanproject get -S tutorials` would use SSH (of course one can always specify a full URL, or indeed `git remote set-url` to change it post-hoc). It does seem wrong to try to make a connection essentially just to parse a string (to determine by `tutorials` whether we "intended" `git@github.com:leanprover-community/tutorials` or not – most users will not have an intention either way). 

For reference:
you should be able to test this issue by moving your private key to `~/.ssh/github_private_keyfile` and then adding the following to `~/.ssh/config`
```
Host github.com
        IdentityFile ~/.ssh/github_private_keyfile
```

PatrickMassot:
This part of `leanproject` is extremely hackish and brittle. I have neither knowledge or time to work on doing it properly, especially since it works well in default setups. PRs are welcome as long as they don't make it more complicated to use `leanproject` for people who have a default setup. 

[...]

PR #74, opened 2020-07-24
SSH stuff

mattearnshaw:
This lets `git` handle all the stuff related to SSH rather than rolling it ourselves, and should work on all setups.

If a full url is not provided, it tries SSH first then falls back to HTTPS if `git` returns permission denied (user has not set up SSH, or SSH keyfile password entered incorrectly 3 times)

```
# User with no SSH keys configured
~> leanproject get tutorials
Cloning from git@github.com:leanprover-community/tutorials.git
SSH permission denied, trying HTTPS...
Cloning from https://github.com/leanprover-community/tutorials.git
```

```
~> leanproject get tutorials
Cloning from git@github.com:leanprover-community/tutorials.git
Enter passphrase for key '/home/user/.ssh/my_github_key':  # this message is from git
```

cf. #72

### DIFF
--- a/mathlibtools/leanproject.py
+++ b/mathlibtools/leanproject.py
@@ -180,7 +180,7 @@ def get_project(name: str, new_branch: bool, directory: str = '') -> None:
                                  cache_url, force_download)
     except GitCommandError as err:
         # if full url is provided, do not retry with HTTPS
-        if ('Permission denied' in err.stderr) and not is_url:
+        if not is_url:
             log.info('SSH permission denied, trying HTTPS...')
             try:
                 name, url, branch, is_url = parse_project_name(original_name, ssh=False)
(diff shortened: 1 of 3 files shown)


mathlibtools/leanproject.py
```diff
@@ -204,7 +179,17 @@ def get_project(name: str, new_branch: bool, directory: str = '') -> None:
         LeanProject.from_git_url(url, directory, branch, new_branch,
                                  cache_url, force_download)
     except GitCommandError as err:
-        handle_exception(err, 'Git command failed')
+        # if full url is provided, do not retry with HTTPS
+        if ('Permission denied' in err.stderr) and not is_url:
```
PatrickMassot, line 183: This is typical USA/UK programming... What about users whose locale is not English?

PatrickMassot, line 183: This is typical USA/UK programming... What about users whose locale is not English?

mattearnshaw: guilty as charged – I don't think the git python API provides a sane way to parse `GitCommandError`, so I will just remove this. The worst that will happen is that if a project name is provided that does not exist (e.g. `leanproject get berfectoid`), it will be retried via HTTPS despite failing because it does not exist. Regardless, the user will see a `Repository not found` error after that.

PR #76, opened 2020-08-15
only check for .xz archives

semorrison:
Don't look on github for olean archives, and don't look for .gz and .bz2 archives anymore.

### DIFF
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -149,7 +149,7 @@ def get_mathlib_archive(rev: str, url:str = '', force: bool = False,
     """
 
     # we check for xz archives first
-    fnames = [rev + '.tar.xz', rev + '.tar.gz', rev + '.tar.bz2']
+    fnames = [rev + '.tar.xz']
     paths = [DOT_MATHLIB/fname for fname in fnames]
     if not force:
         log.info('Looking for local mathlib oleans')
@@ -166,11 +166,6 @@ def get_mathlib_archive(rev: str, url:str = '', force: bool = False,
             return path
         except LeanDownloadError:
             pass
-    log.info('Looking for GitHub mathlib oleans')
-    # nightlies will only store gz archives
-    path = DOT_MATHLIB / (rev + '.tar.gz')
-    download(nightly_url(rev, repo), path)
-    log.info('Found GitHub mathlib oleans')
     return path
 
 def parse_version(version: str) -> VersionTuple:
@@ -747,5 +742,3 @@ class LeanProject:
         """Safe declaration into a pickle file target"""
         with target.open('wb') as f:
             pickle.dump(self.list_decls(), f, pickle.HIGHEST_PROTOCOL)
-
-
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -157,8 +157,8 @@ def get_mathlib_archive(rev: str, url:str = '', force: bool = False,
             return path
     log.info('Looking for remote mathlib oleans')
     base_url = url or get_download_url()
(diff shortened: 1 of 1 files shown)


mathlibtools/lib.py
```diff
@@ -149,7 +149,7 @@ def get_mathlib_archive(rev: str, url:str = '', force: bool = False,
     """
 
     # we check for xz archives first
-    fnames = [rev + '.tar.xz', rev + '.tar.gz', rev + '.tar.bz2']
```
gebner, line 152: @robertylewis Did you also migrate the `tar.gz` archives to `tar.xz`?

robertylewis: Yes, every archive should exist in both forms right now.

mathlibtools/lib.py
```diff
@@ -149,7 +149,7 @@ def get_mathlib_archive(rev: str, url:str = '', force: bool = False,
...
+    fnames = [rev + '.tar.xz']
```
PatrickMassot, line 152: There is no point in keeping a list with only one element, this function must be simplified more drastically.

bryangingechen: I pushed a commit that addressed this.

mathlibtools/lib.py
```diff
@@ -148,29 +148,21 @@ def get_mathlib_archive(rev: str, url:str = '', force: bool = False,
...
-    log.info('Found GitHub mathlib oleans')
+    log.info('Looking for remote mathlib oleans')
+    try:
+        base_url = url or get_download_url()
+        download(base_url+fname, path)
+        log.info('Found mathlib oleans at '+base_url)
+        return path
+    except LeanDownloadError:
+        pass
     return path
```
eric-wieser, line 166:
```suggestion
    base_url = url or get_download_url()
    download(base_url+fname, path)
    log.info('Found mathlib oleans at '+base_url)
    return path
```
since the exception will be caught and handled at https://github.com/leanprover-community/mathlib-tools/blob/78c58aa51445e2833a69d09d1adecc8699a29539/mathlibtools/leanproject.py#L218

mathlibtools/lib.py
```diff
@@ -148,29 +148,21 @@ def get_mathlib_archive(rev: str, url:str = '', force: bool = False,
...
-    download(nightly_url(rev, repo), path)
-    log.info('Found GitHub mathlib oleans')
+    log.info('Looking for remote mathlib oleans')
+    try:
+        base_url = url or get_download_url()
+        download(base_url+fname, path)
+        log.info('Found mathlib oleans at '+base_url)
+        return path
+    except LeanDownloadError:
+        pass
```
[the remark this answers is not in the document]
bryangingechen: Good catch, thanks!

eric-wieser (approved):

mathlibtools/lib.py
```diff
@@ -148,29 +148,17 @@ def get_mathlib_archive(rev: str, url:str = '', force: bool = False,
...
-            pass
-    log.info('Looking for GitHub mathlib oleans')
-    # nightlies will only store gz archives
-    path = DOT_MATHLIB / (rev + '.tar.gz')
-    download(nightly_url(rev, repo), path)
-    log.info('Found GitHub mathlib oleans')
+    log.info('Looking for remote mathlib oleans')
+    base_url = url or get_download_url()
+    download(base_url+fname, path)
+    log.info('Found mathlib oleans at '+base_url)
```
eric-wieser, line 161: PEP8 would say
```suggestion
    download(base_url + fname, path)
    log.info('Found mathlib oleans at ' + base_url)
```

PR #78, opened 2020-09-03
Use appdirs for finding a platform-specific dir for mathlib caches.

Julian:
Hello. Sending this to see if you'd be up for it.

This patch changes the directory previously used to store mathlib caches (`~/.mathlib`) to use [appdirs](https://pypi.org/project/appdirs/), a (simple somewhat widely used) package for finding a platform-specific directory to store things in (config, data, etc.).

E.g. on Linux, it'd instead go in `~/.cache/mathlib`.

Alongside that, I also added a `MATHLIB_CACHE_DIR` environment variable in case anyone wanted to move the directory to some other specific location.

Hopefully the above makes sense, comment welcome of course.

### DIFF
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -14,6 +14,7 @@ from datetime import datetime
 from typing import Iterable, Union, List, Tuple, Optional, Dict
 from tempfile import TemporaryDirectory
 
+import appdirs # type: ignore
 import networkx as nx # type: ignore
 import requests
 from tqdm import tqdm # type: ignore
@@ -69,7 +70,9 @@ def nightly_url(rev: str, proj_repo: Optional[Repo] = None) -> str:
     return asset.browser_download_url
 
 
-DOT_MATHLIB = Path.home()/'.mathlib'
+DOT_MATHLIB = Path(
+    os.environ.get("MATHLIB_CACHE_DIR", appdirs.user_cache_dir("mathlib")),
+)
 AZURE_URL = 'https://oleanstorage.azureedge.net/mathlib/'
 
 DOT_MATHLIB.mkdir(parents=True, exist_ok=True)
--- a/default.nix
+++ b/default.nix
@@ -9,6 +9,6 @@ buildPythonApplication {
   doCheck = false;
 
   propagatedBuildInputs = [
-    PyGithub GitPython toml click tqdm paramiko networkx pydot pyyaml
+    PyGithub GitPython toml click tqdm paramiko networkx pydot pyyaml appdirs
   ];
 }
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,6 @@ setuptools.setup(
         "Operating System :: OS Independent" ],
     python_requires='>=3.5',
     install_requires=['toml>=0.10.0', 'PyGithub', 'certifi', 'gitpython>=2.1.11', 'requests',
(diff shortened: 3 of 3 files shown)


PatrickMassot:
I don't mind, but I'm not really seeing the point either. What problem is this meant to solve? Also, I'd like to know what Windows and MacOS think about this.

Julian:
(I'm on macOS FWIW -- where the directory is ~/Library/ApplicationSupport/mathlib)

The point is quite petty :), it's "mathlib is dirtying my home directory".

But yeah either it's useful or not, so if that's not something anyone else cares about obviously don't mind if it's closed.

robertylewis:
I'm no expert here, but it seems like a reasonable change to me. Best to use as standard a location as possible. 

Maybe a nice addition would be a flag to query `leanproject` about where it's storing archives, and even better a config option to change it, but this might be going too far.

mathlibtools/lib.py
```diff
@@ -69,7 +70,9 @@ def nightly_url(rev: str, proj_repo: Optional[Repo] = None) -> str:
     return asset.browser_download_url
 
 
-DOT_MATHLIB = Path.home()/'.mathlib'
+DOT_MATHLIB = Path(
+    os.environ.get("MATHLIB_CACHE_DIR", appdirs.user_data_dir("mathlib")),
```
gebner, line 74:
```suggestion
    os.environ.get("MATHLIB_CACHE_DIR", appdirs.user_cache_dir("mathlib")),
```

Julian: Yeah, good point. Applying.

Julian:
> I'm no expert here, but it seems like a reasonable change to me. Best to use as standard a location as possible.
> 
> Maybe a nice addition would be a flag to query `leanproject` about where it's storing archives, and even better a config option to change it, but this might be going too far.

These sound reasonable to me too. If they're desirable I'm likely happy to get back to them in a follow-up (probably not immediately though).

gebner:
I don't think anything would break by changing the directory, it's just a cache after all.

But we should tell people that they can now delete `~/.mathlib`:
```
$ du -sh ~/.mathlib
1.9G    /home/gebner/.mathlib
```

Julian:
Are you suggesting implementing some automated migration there to move the directory, or a doc change (or neither?)

gebner:
> Are you suggesting implementing some automated migration there to move the directory, or a doc change (or neither?)

Automatic migration is probably overkill.  I was thinking of a Zulip post that announces the change.

Julian:
Got it! Sounds good!

PatrickMassot:
@Julian I'd ready to merge this but it needs to stay in sync with [documentation](https://github.com/leanprover-community/leanprover-community.github.io/blob/newsite/templates/leanproject.md). And of course this immediately reveal a flaw in this idea: you can no longer simply explain where `leanproject` will put his stuff. 

What's worse than an app that "dirties your home directory" as you wrote? Answer: an app that dirties a random secret place in your home directory. So you would need to at least add a command `leanproject where` (or some other name) that tells you where is the mathlib cache.

Julian:
We probably both have better things to do than go back and forth on this :) -- if you don't want this how about just adding the MATHLIB_CACHE_DIR environment variable, so people who don't want it in their home directory can put it wherever they want?

PatrickMassot:
Ok, let's do that.

PR #79, opened 2020-09-10
 Allow running leanproject via `python -m`

eric-wieser:
This also removes the weirdness from the commit hooks where one python is used to check if python is installed, then a completely different one is actually used.

### DIFF
--- a/mathlibtools/post-checkout
+++ b/mathlibtools/post-checkout
@@ -10,13 +10,13 @@ CHANGED_BRANCH=$3
 if [ "$CHANGED_BRANCH" -eq "1" ]; then
 	if /usr/bin/env python3 -c ""; then
 		echo "Trying to fetch cached olean"
-		leanproject get-cache
+		/usr/bin/env python3 -m leanproject get-cache
 		# If the current project is not called "mathlib", then get a mathlib cache.
 		leanpkg dump | grep --quiet 'name = "mathlib"'
 		if [ $? -ne 0 ]; then
 			leanproject get-mathlib-cache
 		fi
-		leanproject delete-zombies
+		/usr/bin/env python3 -m leanproject delete-zombies
 	else
 		echo "'env python3' failed; not running post-checkout hook"
 	fi
--- a/mathlibtools/post-checkout
+++ b/mathlibtools/post-checkout
@@ -10,13 +10,13 @@ CHANGED_BRANCH=$3
 if [ "$CHANGED_BRANCH" -eq "1" ]; then
 	if /usr/bin/env python3 -c ""; then
 		echo "Trying to fetch cached olean"
-		/usr/bin/env python3 -m leanproject get-cache
+		leanproject get-cache
 		# If the current project is not called "mathlib", then get a mathlib cache.
 		leanpkg dump | grep --quiet 'name = "mathlib"'
 		if [ $? -ne 0 ]; then
 			leanproject get-mathlib-cache
 		fi
-		/usr/bin/env python3 -m leanproject delete-zombies
+		leanproject delete-zombies
 	else
 		echo "'env python3' failed; not running post-checkout hook"
 	fi
--- a/mathlibtools/post-commit
+++ b/mathlibtools/post-commit
@@ -7,7 +7,7 @@ export PATH=$PATH:/usr/local/bin
 # if python3 isn't available, we don't generate an error; just a message
 if /usr/bin/env python3 -c ""; then
 	echo "Caching olean..."
-	/usr/bin/env python3 -m leanproject mk-cache
+	leanproject mk-cache
 else
     echo "'env python3' failed; not running post-commit hook"
 fi
--- a/mathlibtools/post-commit
+++ b/mathlibtools/post-commit
@@ -7,7 +7,7 @@ export PATH=$PATH:/usr/local/bin
 # if python3 isn't available, we don't generate an error; just a message
 if /usr/bin/env python3 -c ""; then
 	echo "Caching olean..."
-	leanproject mk-cache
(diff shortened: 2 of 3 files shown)


mathlibtools/post-checkout
```diff
@@ -10,13 +10,13 @@ CHANGED_BRANCH=$3
 if [ "$CHANGED_BRANCH" -eq "1" ]; then
 	if /usr/bin/env python3 -c ""; then
 		echo "Trying to fetch cached olean"
-		leanproject get-cache
+		/usr/bin/env python3 -m leanproject get-cache
```
Julian, line 13: (This may be the intention), but this will break in the "reverse" case -- where `leanproject` is on `PATH`, but `python3` is not the interpreter it is installed to.

eric-wieser: Correct - but in that situation the test for python3 is already nonsense.

PatrickMassot:
I have nothing to say about the commit hooks, I never used them and have nothing to do with them. But the main modification is really asking for trouble. If your python setup is broken then such a hack is likely to bring you only to more confusion later.

eric-wieser:
> But the main modification is really asking for trouble

The approach of adding a `__main__` to allow `python -m mathlib.leanproject` is standard practice; `python -m http.server` is a builtin example, and `python -m pytest` and `python -m pip` are two third party ones.

I'm happy to revert the hook changed, but I really think the `__main__` addition is a good idea.

mathlibtools/post-checkout
gebner: Concretely, this would break on nixos:
```
$ leanproject help
Try 'leanproject -h' for help.

Error: No such command 'help'.
$ python3 -m leanproject
/run/current-system/sw/bin/python3: No module named leanproject
```

The `env python3` check also looks bogus to me.  Shouldn't it be `command -v leanproject >/dev/null`?  (I don't use the hooks though.)

gebner (changes requested):
The `__name__ == '__main__'` part seems innocuous to me.

mathlibtools/post-checkout
eric-wieser, line 13:
```suggestion
		leanproject get-cache
```

mathlibtools/post-checkout
```diff
@@ -10,13 +10,13 @@ CHANGED_BRANCH=$3
...
 		# If the current project is not called "mathlib", then get a mathlib cache.
 		leanpkg dump | grep --quiet 'name = "mathlib"'
 		if [ $? -ne 0 ]; then
 			leanproject get-mathlib-cache
 		fi
-		leanproject delete-zombies
+		/usr/bin/env python3 -m leanproject delete-zombies
```
eric-wieser, line 19:
```suggestion
		leanproject delete-zombies
```

mathlibtools/post-commit
```diff
@@ -7,7 +7,7 @@ export PATH=$PATH:/usr/local/bin
 # if python3 isn't available, we don't generate an error; just a message
 if /usr/bin/env python3 -c ""; then
 	echo "Caching olean..."
-	leanproject mk-cache
+	/usr/bin/env python3 -m leanproject mk-cache
```
eric-wieser, line 10:
```suggestion
	leanproject mk-cache
```

mathlibtools/post-checkout
eric-wieser: Ah, this line has a typo - it should have been `python3 -m mathlibtools.leanproject`.

eric-wieser:
I've reverted the changes to the hooks. In my opinion it would make sense to entirely drop the bash hooks, and have a single python-entry-point to do all the hook work.

mathlibtools/post-checkout
gebner: That's not the issue, none of the leanproject modules are in the PYTHONPATH even though `leanproject` is installed.
```
$ leanproject help
Try 'leanproject -h' for help.

Error: No such command 'help'.
$ python3 -m mathlibtools.leanproject
/run/current-system/sw/bin/python3: Error while finding module specification for 'mathlibtools.leanproject' (ModuleNotFoundError: No module named 'mathlibtools')
```

eric-wieser: In that case, the issue is that `python3` is not referring to the same version of python as the one in the shebang of `leanproject` (all the more reason not to check for `python3`). What does `head $(which leanproject)` give?

gebner: No, the `python3` is the same in both cases.
```
$ tail -n1 (which leanproject)
exec -a "$0" "/nix/store/21c77l6kirhlngw6gc0ky9r9rlmncll6-mathlib-tools-0.0.9b/bin/.leanproject-wrapped"  "$@"
$ head -n1 /nix/store/21c77l6kirhlngw6gc0ky9r9rlmncll6-mathlib-tools-0.0.9b/bin/.leanproject-wrapped
#!/nix/store/z65l1jqvxa58zzwwa3bvglb6asj4y8cv-python3-3.8.5/bin/python3.8
$ readlink (which python3.8)
/nix/store/z65l1jqvxa58zzwwa3bvglb6asj4y8cv-python3-3.8.5/bin/python3.8
```

However `leanproject` changes the python path before it runs.

eric-wieser: Ok, I see what's happening. The default pip behave is to behave like both the nixos `buildPythonPackage` and `buildPythonApplication` at the same time.

You've presumably installed directly via the nixos method, and not via pip.

eric-wieser:
Is there anything left to do here?

PatrickMassot:
I merged since the current version looks innocuous, even though I still think it's a bad idea to try to use it that way.

PR #86, opened 2020-10-23
Remove Py35, Add Py39, Unify CI

Julian:
Three related changes here:

* Remove Python3.5, which is [EOL'ed](https://www.python.org/dev/peps/pep-0478/#id4)
* Add Python3.9 which was recently released
* Merge all the CI together and run it through tox, such that what someone runs locally matches what will run in GHA.

Comments obviously welcome.

(This started mostly as a roundabout way to address #85 failing on 3.5 which doesn't yet have f-strings, obviously there are simpler fixes to that, but yeah figured you'd likely be OK with these changes.)


PatrickMassot:
Thanks! I hate setting up CI, thank you very much for working on this. The multiple CI files were probably the first thing that worked.

Julian:
Cool, happy to, CI is indeed terrible to setup so all of this is generated by a template I use so that I never have to do it myself.

Can I offer you a CI script to do automatic releases when you push tags? It's way more convenient to push to PyPI than the other nonsense locally.

I especially will offer you that if you in exchange push a release :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
...
+      with:
+        python-version: ${{ matrix.python-version.name }}
+    - name: Ensure we have new enough versions to respect python_version
+      run: python -m pip install -U pip
+    - name: Install dependencies
+      run: brew install gmp coreutils
+      if: runner.os == 'macOS'
+    - name: Install elan
+      run: |
+          curl https://raw.githubusercontent.com/Kha/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain leanprover-community/lean:3.5.1 -y
```
Julian, line 41: One more random thing I noticed while doing this -- you're running on lean:3.5.1, obviously if you want we could matrix build across say 2 lean versions, the oldest one this supports and then the newest release.

PR #88, opened 2020-12-02
Fix leanproject import-graph to not produce a graph with no edges

eric-wieser:
`src_directory` (on my machine) is a relative path, `lean --path` outputs absolute paths, and `relative_to` fails every time for `abs_path.relative_to(rel_path)`.

Tested on windows.

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Diagnosing.20cyclic.20imports/near/218610115)


eric-wieser:
CI failure looks unrelated.

gebner:
So github finally disabled `add-path`.  #89

PR #92, opened 2021-01-21
feat(leanproject) add a function to list imports that can be removed due to being imported transitively

alexjbest:
`leanproject reduce-imports --file foo.bar` will list imports in `foo/bar.lean` that are imported by other imports of `foo.bar` and could therefore be removed, `--sed` will print out a sed script to cut out those lines.

### DIFF
--- a/mathlibtools/leanproject.py
+++ b/mathlibtools/leanproject.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from datetime import datetime
 from typing import Tuple, Optional
 from getpass import getpass
+import networkx as nx # type: ignore
 
 from git.exc import GitCommandError # type: ignore
 
@@ -287,6 +288,20 @@ def global_upgrade() -> None:
     proj = LeanProject.user_wide(cache_url, force_download)
     proj.upgrade_mathlib()
 
+def _import_graph_(to: str = None, from_: str = None) -> nx.DiGraph:
+    project = proj()
+    graph = project.import_graph
+    if to and from_:
+        G = graph.path(start=from_, end=to)
+    elif to:
+        G = graph.ancestors(to)
+    elif from_:
+        G = graph.descendants(from_)
+    else:
+        G = graph
+    return G
+
+
 @cli.command()
 @click.option('--to', 'to', default=None,
               help='Return only imports leading to this file.')
@@ -303,17 +318,39 @@ def import_graph(to: Optional[str], from_: Optional[str], output: str) -> None:
     By default the graph will be written to 'import_graph.dot'.
     For .dot, .pdf, .svg, or .png output you will need to install 'graphviz' first.
     """
-    project = proj()
-    graph = project.import_graph
-    if to and from_:
-        G = graph.path(start=from_, end=to)
-    elif to:
-        G = graph.ancestors(to)
-    elif from_:
-        G = graph.descendants(from_)
-    else:
-        G = graph
+    G = _import_graph_(to, from_)
     G.write(Path(output))
+
+
+@cli.command()
+@click.option('--sed', 'sed', default=False, is_flag=True,
+              help='Instead of printing a list of removable imports, print a sed script that can be run to remove the imports.')
+@click.argument('file', default=None, required=False)
+def reduce_imports(file: str, sed: bool = False) -> None:
+    """List imports that can be removed in the project.
+
+    Argument '--file' should be specified as a
+    Lean import (e.g. 'data.mv_polynomial') rather than a file name.
+    """
+    G = _import_graph_(to = file)
+    H = nx.transitive_reduction(G)
+    if file:
+        fs = [file]
+    else:
+        fs = G.nodes
+    for f in fs:
+        if f == "all":
+            continue
+        Gf = [e for e in G.edges if e[1] == f]
+        Hf = [e for e in H.edges if e[1] == f]
+        o = [e for e in Gf if e[1] == f and e not in H.edges]
+        if sed:
+            for df in o:
+                # probably not the right command on osx
+                print("sed -i '/^import {line}$/d' src/{file}.lean".format(file=df[1].replace(".","/"), line=df[0]))
+        else:
+            if o:
+                print(o)
 
 
 @cli.command()
--- a/mathlibtools/leanproject.py
+++ b/mathlibtools/leanproject.py
@@ -3,7 +3,6 @@ from pathlib import Path
 from datetime import datetime
 from typing import Tuple, Optional
 from getpass import getpass
-import networkx as nx # type: ignore
 
 from git.exc import GitCommandError # type: ignore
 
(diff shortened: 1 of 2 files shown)


mathlibtools/leanproject.py
```diff
@@ -303,19 +318,41 @@ def import_graph(to: Optional[str], from_: Optional[str], output: str) -> None:
...
-        G = graph.descendants(from_)
-    else:
-        G = graph
+    G = _import_graph_(to, from_)
     G.write(Path(output))
 
 
+@cli.command()
+@click.option('--sed', 'sed', default=False, is_flag=True,
+              help='Instead of printing a list of removable imports, print a sed script that can be run to remove the imports.')
```
eric-wieser, line 327: Emitting a git patch file might make the most sense, as almost all users of lean will have git installed.

alexjbest: Agreed, it's probably a little more work though to determine the file offsets? Maybe the best thing to do is make the replacement using python itself

alexjbest: Actually probably grep can do what sed does here even

PatrickMassot:
For some reason I completely missed that PR. I'm discovering while investigating some other feature request. I think it's a nice addition, but it's not done in the correct place. The idea is that all the actual work is done in `lib.py` and `leanproject.py` is only about the command line interface.

mathlibtools/lib.py
```diff
@@ -634,6 +634,32 @@ def import_graph(self) -> 'ImportGraph':
         self._import_graph = G
         return G
 
+    def reduce_imports(self, file: str, sed: bool = False) -> Iterable[Any]:
```
PatrickMassot, line 637: The return type annotation is too vague.

eric-wieser: I think it makes sense to split this into two functions, a non-sed version and a sed version that calls the non-sed version.

PatrickMassot: That would probably be clearer indeed.

alexjbest: Agreed

mathlibtools/lib.py
```diff
@@ -639,6 +639,33 @@ def import_graph(self) -> 'ImportGraph':
         self._import_graph = G
         return G
 
+    def reduce_imports(self, file: str) -> Iterable[List[Any]]:
```
PatrickMassot, line 642: I don't understand how the type can be mysterious here. Why `Any`?

alexjbest: You are right (of course!), it was simply my inexperience with typing made me think that as networkx doesn't have type annotations this was the best possible.

mathlibtools/lib.py
```diff
@@ -639,6 +639,33 @@ def import_graph(self) -> 'ImportGraph':
         self._import_graph = G
         return G
 
+    def reduce_imports(self, file: str) -> Iterable[List[Tuple[str,str]]]:
```
eric-wieser, line 642: Can you add a docstring indicating what this type contains? I assume the `Tuple[str,str]` are edges (in which direction?), but I have no idea what the intermediate list structure is.

eric-wieser:
```suggestion
    def reduce_imports(self, file: str) -> Iterable[List[Tuple[str, str]]]:
```

alexjbest: Done

mathlibtools/lib.py
```diff
@@ -639,6 +639,38 @@ def import_graph(self) -> 'ImportGraph':
         self._import_graph = G
         return G
 
+    def reduce_imports(self, file: str) -> Iterable[List[Tuple[str, str]]]:
+        """
+        An iterator over files with removable imports, for each file yielding
+        a list of removable imports in the format
+        `("removable.import", "source.file")`.
```
eric-wieser, line 646: This is a weird representation. I'd suggest either:
```python
[('src1', ['removable1', 'removable2']),
 ('src2', ['removable3', 'removable4', 'removable5'])]
```
or
```python
[('removable1', 'src1'),
 ('removable2', 'src1'),
 ('removable3', 'src2'),
 ('removable4', 'src2'),
 ('removable5', 'src2')]
```
rather than than the weird structured yet also duplicated representation you've gone for,
```python
[[('removable1', 'src1'),
  ('removable2', 'src1')],
 [('removable3', 'src2'),
  ('removable4', 'src2'),
  ('removable5', 'src2')]]
```

mathlibtools/lib.py
```diff
@@ -639,6 +639,38 @@ def import_graph(self) -> 'ImportGraph':
...
+        else:
+            fs = G.nodes
+        for f in fs:
+            if f == "all":
+                continue
+            Gf = [e for e in G.edges if e[1] == f]
+            Hf = [e for e in H.edges if e[1] == f]
+            o = [e for e in Gf if e[1] == f and e not in H.edges]
+            if o:
+                yield o
```
eric-wieser, line 666:
```suggestion
            o = [e[0] for e in Gf if e[1] == f and e not in H.edges]
            if o:
                yield (f, o)
```

mathlibtools/lib.py
```diff
@@ -639,6 +639,38 @@ def import_graph(self) -> 'ImportGraph':
...
+        H = nx.transitive_reduction(G)
+        if file:
+            fs = [file]
+        else:
+            fs = G.nodes
+        for f in fs:
+            if f == "all":
+                continue
+            Gf = [e for e in G.edges if e[1] == f]
+            Hf = [e for e in H.edges if e[1] == f]
```
eric-wieser, line 663: `Hf` is unused?

mathlibtools/lib.py
```diff
@@ -639,6 +639,38 @@ def import_graph(self) -> 'ImportGraph':
...
+            o = [e for e in Gf if e[1] == f and e not in H.edges]
+            if o:
+                yield o
+
+    def reduce_imports_sed(self, file: str) -> Iterable[str]:
+        for o in self.reduce_imports(file):
+            for df in o:
+                # probably not the right command on osx
+                yield "sed -i '/^import {line}$/d' src/{file}.lean".format(file=df[1].replace(".","/"), line=df[0])
```
eric-wieser, line 672:
```suggestion
        for src, removable in self.reduce_imports(file):
            for r in removable:
                # probably not the right command on osx
                yield "sed -i '/^import {line}$/d' src/{file}.lean".format(file=src.replace(".","/"), line=r)
```

mathlibtools/lib.py
alexjbest: Also fine by me, the internal representation is simply the output of networkx rather than a deliberate choice.

mathlibtools/lib.py
alexjbest: Thanks, this should have been referenced in the next line.

mathlibtools/lib.py
alexjbest: Thanks!

mathlibtools/lib.py
```diff
@@ -639,6 +639,38 @@ def import_graph(self) -> 'ImportGraph':
         self._import_graph = G
         return G
 
+    def reduce_imports(self, file: str) -> Iterable[Tuple[str, List[str]]]:
+        """
+        An iterator over files with removable imports, for each file yielding
+        a list of removable imports in the format
+        `("source.file", ["removable.import", "another.removable.import"])`.
+        """
+        # Importing networkx slow, so don't do it until this function
```
PatrickMassot, line 648:
```suggestion
        # Importing networkx is slow, so don't do it until this function
```

PR #93, opened 2021-01-25
fix(install_macos.sh): remove deprecated brew cask install

fpvandoorn:
See also https://github.com/Homebrew/discussions/discussions/340

---

@rwbarton and I helped some people to install Lean on their computers. We ran into two problems:

* One error `Calling brew cask install is disabled! Use brew install [--cask] instead.`
The script worked after replacing `brew cask install visual-studio-code` by `brew install --cask visual-studio-code`
See also e.g. https://github.com/Homebrew/discussions/discussions/340

* The other error was that `leanpkg` and `elan` were not in the path after installation, but I think that problem has the same cause. We didn't run `source ~/.profile` after correctly installing mathlibtools.

@Vierkantor @PatrickMassot

### DIFF
--- a/scripts/install_macos.sh
+++ b/scripts/install_macos.sh
@@ -22,6 +22,6 @@ pipx install mathlibtools
 
 # Install and configure VS Code
 if ! which code; then
-brew cask install visual-studio-code
+brew install --cask visual-studio-code
 fi
 code --install-extension jroesch.lean


Vierkantor (approved):
Looks good to me. I don't have a machine to test the full script at the moment though.

Julian:
Presumably would be nice to have CI run the script.

I put a branch here that does that: https://github.com/Julian/mathlib-tools/tree/ci-for-install-macos which indeed fails without this commit and passes with it (technically it's not really a pristine CI, since GHA runners install homebrew before executing actions, whereas this script appears to want to help folks install homebrew, but yeah possibly better than nothing). Obviously feel free to take that branch.

jasdev:
> Looks good to me. I don't have a machine to test the full script at the moment though.

Just ran into this snag on a fresh install and it went through cleanly after @fpvandoorn’s update (which I accidentally duped in https://github.com/leanprover-community/mathlib-tools/pull/98 hah 😄)

bryangingechen:
I'm going to merge this now as it seems straightforward and there's no reason to force more people to run into this.

@Julian do you mind PRing some version of your CI script?

PR #96, opened 2021-02-23
Do not touch the working tree when running `leanproject get-cache`

eric-wieser:
If I run get-cache when working on mathlib, and have a dirty working directory, it requires me to add `--force`.
This is probably sensible, because if I do this I'm likely to throw away whatever local cache I have, and suffer through a long build.

However, when I add `--force`, it blows away not only my local cache, but my local changes to `lean` files too.

This behavior is clearly undesirable when used in conjunction with `--rev`: `leanproject get-cache --rev old --force` resets my local files to `HEAD` but my oleans to `old`. The expected behavior is to do only the latter.

This PR also makes it unnecessary to pass `--force` when passing `--rev`, as the only risk associated with dirty changes is also associated with asking for a non-head revision.

### DIFF
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -451,7 +451,7 @@ class LeanProject:
         self.clean_mathlib()
         self.mathlib_folder.mkdir(parents=True, exist_ok=True)
         unpack_archive(archive, self.mathlib_folder)
-        if rev:
+        if rev or self.is_dirty:
             self.delete_zombies()
         # Let's now touch oleans, just in case
         touch_oleans(self.mathlib_folder)
@@ -478,8 +478,11 @@ class LeanProject:
         """
         if not self.repo:
             raise LeanProjectError('This project has no git repository.')
-        if self.is_dirty and not force:
-            raise LeanDirtyRepo('Cannot get cache for a dirty repository.')
+        if self.is_dirty and rev is None and not force:
+            # if the user passed `rev`, they deliberately want a cache that does
+            # not match their local changes anyway.
+            raise LeanDirtyRepo('Getting the cache may result in a '
+                                'time-consuming rebuild.')
         if self.is_mathlib:
             self.get_mathlib_olean(rev)
         else:
@@ -583,9 +586,8 @@ class LeanProject:
     def clean_mathlib(self, force: bool = False) -> None:
         """Restore git sanity in mathlib"""
         if self.is_mathlib:
-            if not self.is_dirty or force:
-                assert self.repo
-                self.repo.head.reset(working_tree=True)
+            # the mathlib folder contains user edited files - do not touch it!
+            pass
         elif self.mathlib_folder.exists():
             mathlib = Repo(self.mathlib_folder)
             mathlib.head.reset(working_tree=True)


eric-wieser:
[Zulip](https://leanprover.zulipchat.com/#narrow/stream/116290-rss/topic/Recent.20Commits.20to.20mathlib-tools.3Amaster/near/220862021)

eric-wieser:
CI failure is a network issue

pechersky:
Is it possible to make a similar change for `leanproject build`? If I have uncommited changes, those are blown away. It makes sense to me that `leanproject build` was just a wrapper for `lean --make src`.

eric-wieser:
This change should have applied to `leanproject build` too

bryangingechen:
I reran the workflows to give this a green checkmark.

Would it be possible to add a test to https://github.com/leanprover-community/mathlib-tools/blob/master/tests/test_functional.py ?

eric-wieser:
It would, but I'm not sure exactly what the test would be, nor am I super motivated to write some right now.

bryangingechen:
Fair enough. I might take a stab at tests later this week then, given that I suspect that will make Patrick's job a little easier.

eric-wieser:
[Another Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general)

mathlibtools/lib.py
```diff
@@ -478,8 +478,11 @@ def get_cache(self, rev: Optional[str] = None, force: bool = False) -> None:
         """
         if not self.repo:
             raise LeanProjectError('This project has no git repository.')
-        if self.is_dirty and not force:
-            raise LeanDirtyRepo('Cannot get cache for a dirty repository.')
+        if self.is_dirty and rev is None and not force:
+            # if the user passed `rev`, they deliberately want a cache that does
+            # not match their local changes anyway.
+            raise LeanDirtyRepo('Getting the cache may result in a '
+                                'time-consuming rebuild.')
```
PatrickMassot, line 485: I don't understand this error message. It looks like a warning to the user and will still stop everything.

mathlibtools/lib.py
```diff
@@ -583,9 +586,8 @@ def build(self) -> None:
     def clean_mathlib(self, force: bool = False) -> None:
         """Restore git sanity in mathlib"""
```
PatrickMassot, line 587: If you really want this function to do nothing while working on mathlib then it should be renamed to `clean_mathlib_dep` or something like this. I still don't understand your workflow, but I wouldn't mind merging this change.

mathlibtools/lib.py
```diff
@@ -583,9 +586,8 @@ def build(self) -> None:
...
         if self.is_mathlib:
-            if not self.is_dirty or force:
-                assert self.repo
-                self.repo.head.reset(working_tree=True)
+            # the mathlib folder contains user edited files - do not touch it!
+            pass
```
PatrickMassot, line 590: Why doing a `pass` rather than `return`? Isn't it only obfuscating things? And the comment is also very misleading, nothing tells you the folder contains user edited file here.

PatrickMassot (commented):
It may make more sense to keep the cleaning function as it is but calling it only when working on a project which is not mathlib.

mathlibtools/lib.py
eric-wieser: No preference between `pass` and `return` here, either sounds good to me. How about this as the comment?
```
# if this is mathlib itself then the lean files are part of the working tree - do not touch them!
```

mathlibtools/lib.py
eric-wieser: Right - the point is that the user is about to blow away any local olean files they might have from modifications, and once they do that the only way to get them back is to rebuild (at least, I think that's how this works).

So what I'm trying to convey is "if you do this by accident you may end up losing time, pass `--force` to proceed".

I see now that this code isn't quite right either. There are two different types of caches:

* `olean`-only caches, like those uploaded to azure for mathlib
* `lean` + `olean` caches, like those made with `mk-cache`

In the first case, `--force` is really just about being willing to wait a bit longer. In the second case, `--force` can actually destroy changes in your working tree.

PR #97, opened 2021-02-25
fix(mk_cache): make xz archive instead of bz2

robertylewis:
Running `leanproject mk_cache` followed by `leanproject get_cache` in a mathlib repo failed, since `get_cache` only looked for an `xz` file. IIRC we deprecated all other formats. It makes sense for `mk_cache` to use the same format as mathlib CI. This will probably cause minor breakage on upgrade (if anyone actually uses this) but nothing too big.

### DIFF
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -82,7 +82,7 @@ if not DOWNLOAD_URL_FILE.exists():
     set_download_url()
 
 def pack(root: Path, srcs: Iterable[Path], target: Path) -> None:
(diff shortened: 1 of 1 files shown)


PatrickMassot:
Did you check that works? I'm a bit surprised you don't need to patch the `pack` function.

robertylewis:
I didn't test it, no. I thought I did a find/replace for `bz2` -> `xz` but apparently not. Sorry about that, good catch.

bryangingechen:
How hard is it to add a test for the previous failure following what's in `tests/`?

PR #98, opened 2021-03-06
Use --cask flag instead of cask command.

jasdev:
Folding in a [change from Homebrew 2.6.0+](https://brew.sh/2020/12/01/homebrew-2.6.0/), which caused the macOS script to bottom out when trying to install VS Code.

> All `brew cask` commands have been deprecated in favor of `brew` commands (with `--cask`) when necessary


jasdev:
Ah, whoops — realized this is a dupe of https://github.com/leanprover-community/mathlib-tools/pull/93 — sorry about that!

PR #100, opened 2021-04-17
update elan URL

bryangingechen:


.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       if: runner.os == 'macOS'
     - name: Install elan
       run: |
-          curl https://raw.githubusercontent.com/Kha/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain leanprover-community/lean:3.5.1 -y 
+          curl https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain leanprover-community/lean:3.5.1 -y
```
bryangingechen, line 41: This hardcodes Lean 3.5.1; I guess it hasn't hurt up to now, but this might be something we want to change.

Julian:
lgtm if my vote counts

PatrickMassot:
I completely missed that PR. I have no idea where the hardcoded 3.5.1 comes from. What happens if we simply don't specify a default toolchain?

.github/workflows/ci.yml
bryangingechen:
```suggestion
          curl https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain none -y
```

bryangingechen:
I just changed the default toolchain to `none` to follow mathlib's CI; let's see if it works.

.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       if: runner.os == 'macOS'
     - name: Install elan
       run: |
-          curl https://raw.githubusercontent.com/Kha/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain leanprover-community/lean:3.5.1 -y 
+          curl https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain none -y
```
bryangingechen, line 41:
```suggestion
          curl https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh -sSf | sh -s -- -y
```

bryangingechen:
OK, so `none` doesn't work, but not specifying a default works after all!

eric-wieser (approved):
CI passes, so this looks good to me

PR #101, opened 2021-05-03
feat(leanproject): create leanproject.exe for Windows on push

hmonroe:
Create a new workflow pyinstaller-windows.yml that builds a standalone file leanproject.exe whenever new code is pushed. This will allow leanproject to be run on a computer without Python, simplifying installation procedure https://leanprover-community.github.io/install/windows.html


mathlibtools/leanproject.spec
```diff
@@ -0,0 +1,34 @@
+# -*- mode: python ; coding: utf-8 -*-
+
+
+block_cipher = None
+
+
+a = Analysis(['leanproject.py'],
+             pathex=['C:\\Users\\hmonr\\mathlibtools\\mathlib-tools\\mathlibtools'],
```
eric-wieser, line 8: This doesn't look right to me; this path contains your windows username! I don't yet know enough about pyinstaller to suggest what should be here instead though.

hmonroe: Based on my testing, it is ok to delete that line "pathex=['C:\\Users\\hmonr\\mathlibtools\\mathlib-tools\\mathlibtools']". 

So what is the next step?

.github/workflows/main.yml
```diff
@@ -0,0 +1,4 @@
+- name: PyInstaller Windows
+  uses: JackMcKew/pyinstaller-action-windows@main
+    with:
+      path: mathlibtools
```
eric-wieser, line 4: CI is failing on this file. I assume you never meant to commit it anyway?

.github/workflows/pyinstaller-windows.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Package Application
+      uses: JackMcKew/pyinstaller-action-windows@main
+      with:
+        path: mathlibtools
+        spec: leanproject.spec
```
eric-wieser, line 21: Let's make it clear that these requirements are what are used for pyinstaller
```suggestion
        spec: leanproject.spec
        requirements: pyinstaller-requirements.txt
```
(and rename the requirements file below)

.github/workflows/main.yml
hmonroe: Apologies, the instructions said main.yml file is necessary, but I checked and it is not. Fine of course to rename the requirements file.

hmonroe: Or I misread the instructions. I see the example he provided does not include a main.yml file.

.github/workflows/pyinstaller-windows.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: leanproject.exe
```
Julian, line 25: Perhaps it's reasonable enough here to add a second step that uses the artifact that was built in this first step to do a simple smoke test on a Windows machine?

I.e. just run `leanproject.exe --help` or something?

mathlibtools/leanproject.py
```diff
@@ -3,7 +3,6 @@
 from datetime import datetime
 from typing import Tuple, Optional
 from getpass import getpass
-
```
eric-wieser, line 6: This blank line is intentional, it separates the stdlib imports from third-part ones

.github/workflows/pyinstaller-windows.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    - name: Package Application
+      uses: JackMcKew/pyinstaller-action-windows@main
+      with:
+        path: mathlibtools
+        spec: leanproject.spec
+        requirements: pyinstaller-requirements.txt
+      
+    - name: Test .exe file
+      run: dir
```
eric-wieser, line 25:
```suggestion
      run: mathlibtools/dist/windows/leanproject.exe --help
```

.github/workflows/pyinstaller-windows.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    - name: Package Application
+      uses: JackMcKew/pyinstaller-action-windows@main
+      with:
+        path: mathlibtools
+        spec: leanproject.spec
+        requirements: pyinstaller-requirements.txt
+      
+    - name: Test .exe file
+      run: cd mathlibtools
```
eric-wieser, line 25:
```suggestion
```

.github/workflows/pyinstaller-windows.yml
eric-wieser: Ah, the obvious problem here is that we're in a linux container!

eric-wieser:
@hmonroe, I think you're misunderstanding what's going on here. The version I pushed was correct, but the command line is running on ubuntu, and ubuntu can't run exe files.

hmonroe:
Apologies you are being bombarded with pull requests which I was trying to work independently on my own fork. I will try what you suggest but I was trying to rule that we were still in the Wine emulator--the "dir" command worked.

Julian:
At least for me I don't mind :) hopefully I didn't send you on a goose chase setting up the second job, if you need help just speak up!

eric-wieser:
> Apologies you are being bombarded with pull requests which I was trying to work independently on my own fork.

The reason we're seeing these commits is because you create the PR from your master branch rather than creating a branch for just this feature. Either way, we're only being notified because we've commented on the PR, most of the lean community are not. I just wanted to stop you going down the wrong rabbit hole to save you time :)

> I will try what you suggest but I was trying to rule that we were still in the Wine emulator--the "dir" command worked.

`dir` is also [part of "gnu coreutils"](https://www.gnu.org/software/coreutils/manual/html_node/dir-invocation.html), which is the one you are seeing.

hmonroe:
Could one of you please implement the "leanproject.exe --help" change? Another approach would be to fork JackMcKew/pyinstaller-action-windows@main and edit entrypoint.sh to run the leanproject.exe command in the Wine environment already set up there. I have spent several hours on this and have run out of steam.

Julian:
I'm happy to give it a shot but it will likely take me a few days to get to, I've got a few other things on my yak stack.

Julian:
@hmonroe sorry for the delay here, but here's a branch with 3 commits that do the above: https://github.com/Julian/mathlib-tools/tree/windows. Feel free to merge it into yours. (There's a mypy related CI failure I see which has nothing to do with this PR, so we may need to fix that)

The piece you were missing (which I also fought with for a bit to get precisely working) was using `${{ github.workspace }}`, because powershell is being run from some random temporary directory, so the path to relative files isn't the right one. I also ditched using the action you found in favor of just running PyInstaller manually, it's the same amount of work to do so.

I don't have access to a Windows machine to vet the `leanproject.exe` fully, but running `--help` at least is working in CI. If you do, probably good to double check it manually (and if it doesn't work, for us to extend the smoke test).

But I've gotten a build to pass, see e.g.: https://github.com/Julian/mathlib-tools/actions/runs/942808488 (where you'll find a built leanproject.exe).

hmonroe:
@eric-wieser Julian's smoke test looks good. See the section "Check the resulting exe functions" which shows the output of leanproject.exe -h. When I put in the incorrect exe name leanprojectTESTTOBREAK.exe, that triggered a failure of the smoke test. I think leanproject.exe is ready to distribute. There is an unrelated mypy error in CI build.

.github/workflows/pyinstaller-windows.yml
```diff
@@ -0,0 +1,34 @@
...
+        python-version: 3.9
+
+    - name: Install pyinstaller
+      run: python -m pip install pyinstaller -r mathlibtools/pyinstaller-requirements.txt
+
+    - name: Run pyinstaller
+      run: python -m PyInstaller mathlibtools/leanproject.spec
+
+    - name: Check the resulting exe functions
+      run: ${{ github.workspace }}/dist/leanproject.exe --help
```
eric-wieser, line 29: I'm quite worried the exe only works if in this specific location. Have we tested that it works when downloaded (on a system with no installation of mathlib-tools)?

Julian: I have not which is what I meant by "someone with a windows machine should try downloading that".

Definitely agree it's worth doing before merging, I have the same sneaky suspicion given how magical pyinstaller is.

hmonroe: Yes I tested that the download works on a system with no installation ever of mathlib-tools.

hmonroe: Where does this stand? Is there any obstacle to accepting this pull request?

eric-wieser: I'll try to remember to test this tomorrow. The config looks suspect to me still, but that's only a first impression having never used pyinstaller before.

hmonroe: I think you are remembering that the .spec file included a pathname on my PC which I fixed on May 4. https://github.com/leanprover-community/mathlib-tools/pull/101/commits/6f75794175093f5ddcf0b8b9ac637fd2e2e48659 That was just a relic of how the file was generated.

eric-wieser: I tried it out, and after trying a few argument variants, windows defender tried to quarantine it as malware. Unless pyinstaller itself is compromised I assume it's a false alarm.

hmonroe: Yes, that will happen to the first so many people who download it. I have been involved with a distributed computing project (http://vdwnumbers.org/) for ten years--this happens for a few days every time we upgrade the exe files, until enough people click on it as OK. Chrome says "leanproject.exe.zip is not commonly downloaded and may be dangerous.

hmonroe: What is the status of this pull request?

mathlibtools/leanproject.py
```diff
@@ -3,7 +3,6 @@
...
 from git.exc import GitCommandError # type: ignore
```
eric-wieser, line 6:
```suggestion

from git.exc import GitCommandError # type: ignore
```

eric-wieser: At the bottom of this file there is an `if __name__ == "__main__"`. Can you add a comment there to indicate that this is also needed for PyInstaller?

mathlibtools/leanproject.spec
```diff
@@ -0,0 +1,33 @@
+# -*- mode: python ; coding: utf-8 -*-
```
eric-wieser, line 1:
```suggestion
"""
# -*- mode: python ; coding: utf-8 -*-
"""
Originally generated with `pyinstaller`.

In future, we may want to use the approach described at
https://github.com/pyinstaller/pyinstaller/wiki/Recipe-Setuptools-Entry-Point, if we end up adding more entrypoints.
"""
```

mathlibtools/leanproject.spec
```diff
@@ -0,0 +1,39 @@
...
+"""
+Originally generated with `pyinstaller`.
```
eric-wieser, line 3: @hmonroe, do you remember the exact command you used to generate this? If so, can you edit this line to describe it?

hmonroe: Yes I added the command to generate the .spec file

hmonroe:
So this PR can be approved now?

PR #102, opened 2021-05-16
Simplify the Install Script on macOS to use brew

Julian:
With `elan` and `mathlibtools` itself in Homebrew now, we can cut some of this down.

### DIFF
--- a/scripts/install_macos.sh
+++ b/scripts/install_macos.sh
@@ -1,27 +1,21 @@
 #!/bin/bash
 # Install Homebrew
-/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+set -e
 
-# Install dependencies
-brew install gmp coreutils python3 pipx
-
-# Install Elan
-# At startup, Bash sources .bash_profile, or if that doesn't exist, .profile.
-# Elan adds itself to the PATH in .bash_profile, or if that doesn't exist, .profile.
-# pipx only adds itself to the PATH in .bash_profile.
-# So we will create .bash_profile if it doesn't exist yet, ensuring .profile still gets loaded after installing pipx.
-if ! [ -r ~/.bash_profile ]; then
-echo '[ -r ~/.profile ] && source ~/.profile' >> ~/.bash_profile
+if ! which brew > /dev/null; then
+    # Install Homebrew
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+else
+    # Update it, in case it has been ages since it's been updated
+    brew update
 fi
-curl https://raw.githubusercontent.com/Kha/elan/master/elan-init.sh -sSf | sh
 
-# Install mathlib supporting tools
-pipx ensurepath
-source ~/.bash_profile
-pipx install mathlibtools
+brew install elan mathlibtools
+elan toolchain install stable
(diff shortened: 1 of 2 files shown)


scripts/install_macos.sh
```diff
@@ -1,24 +1,12 @@
...
-# Install Elan
-# At startup, Bash sources .bash_profile, or if that doesn't exist, .profile.
-# Elan adds itself to the PATH in .bash_profile, or if that doesn't exist, .profile.
-# pipx only adds itself to the PATH in .bash_profile.
-# So we will create .bash_profile if it doesn't exist yet, ensuring .profile still gets loaded after installing pipx.
-if ! [ -r ~/.bash_profile ]; then
-echo '[ -r ~/.profile ] && source ~/.profile' >> ~/.bash_profile
-fi
-curl https://raw.githubusercontent.com/Kha/elan/master/elan-init.sh -sSf | sh
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```
carlocab, line 5: Can we wrap this in an `if ! which brew` just like the check for `code` below? I'm not sure the install script is idempotent.

Julian: Good idea, yep, done!

PatrickMassot:
In case this isn't clear: I have no opinion about this MacOS stuff, so don't expect a review from me. But I'd be happy to hit the merge button if a couple of Mac users can test this and report success.

scripts/install_macos.sh
```diff
@@ -1,27 +1,21 @@
...
-brew install gmp coreutils python3 pipx
-
-# Install Elan
-# At startup, Bash sources .bash_profile, or if that doesn't exist, .profile.
-# Elan adds itself to the PATH in .bash_profile, or if that doesn't exist, .profile.
-# pipx only adds itself to the PATH in .bash_profile.
-# So we will create .bash_profile if it doesn't exist yet, ensuring .profile still gets loaded after installing pipx.
-if ! [ -r ~/.bash_profile ]; then
-echo '[ -r ~/.profile ] && source ~/.profile' >> ~/.bash_profile
+if ! which brew; then
```
carlocab, line 5:
```suggestion
if ! which brew > /dev/null; then
```
Without this, it'll print output to stdout, which we probably don't need to see.

scripts/install_macos.sh
```diff
@@ -1,27 +1,21 @@
...
-# Install mathlib supporting tools
-pipx ensurepath
-source ~/.bash_profile
-pipx install mathlibtools
+brew install elan mathlibtools
+elan toolchain install stable
+elan default stable
 
 # Install and configure VS Code
 if ! which code; then
```
carlocab, line 18:
```suggestion
if ! which code > /dev/null; then
```
Same comment.

carlocab:
> In case this isn't clear: I have no opinion about this MacOS stuff, so don't expect a review from me. But I'd be happy to hit the merge button if a couple of Mac users can test this and report success.

👋 Mac user here. This works for me.

Julian:
Thanks! Another good point, fixed.

Julian:
@semorrison any chance you feel comfortable +1'ing this as well?

semorrison:
Okay, I'm game! @PatrickMassot, let's merge.

fpvandoorn:
Looks like there is enough support

bors merge

PR #103, opened 2021-05-17
Update macOS installer

carlocab:
Everything is available from Homebrew now, so it seems simpler to
install everything using that. I've omitted the dependencies (gmp,
coreutils and python3), as brew should install those too if they're not
available. (I don't mind putting them back too if you prefer.)

Also, check that there is no `brew` command before trying to run the
Homebrew install script, as I'm not actually certain the script is
idempotent.


carlocab:
Oops. Should've checked the open PRs.

Duplicate of #102

PR #104, opened 2021-07-13
feat(leanproject): download to a tempfile

Vierkantor:
If leanproject is downloading a file (e.g. mathlib oleans) and gets interrupted, we should not keep around the broken archive file. Otherwise, if you re-run the command, leanproject will think the partial download is in fact a complete archive and it will try to extract a broken file.

This PR fixes that issue by creating a temporary file in the `download` function and moving this over the target file when the download has completed successfully.

I have successfully tested the changes on my Linux computer by running `leanproject get-cache` and interrupting the download, then running `leanproject get-cache` again and verifying the oleans are available. Since atomic operations are OS-specific, someone with access to a Windows computer should definitely test this PR as well.
(Is there a way to insert a connection error in the middle of a download automatically, so we can add this functionality to the `tox` tests?)

### DIFF
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -11,7 +11,7 @@ import platform
 import subprocess
 import pickle
 from datetime import datetime
-from typing import Iterable, Union, List, Tuple, Optional, Dict, TYPE_CHECKING
+from typing import Iterable, IO, Union, List, Tuple, Optional, Dict, TYPE_CHECKING
 from tempfile import TemporaryDirectory
 
 import requests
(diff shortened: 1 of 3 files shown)


mathlibtools/lib.py
```diff
@@ -117,14 +117,19 @@ def download(url: str, target: Path) -> None:
...
     progress.close()
     if total_size != 0 and progress.n != total_size:
         raise LeanDownloadError('Failed to download ' + url)
 
+def download(url: str, target: Path) -> None:
+    """Download from url into the target path"""
+    with tempfile.NamedTemporaryFile(mode='wb') as temp:
+        log.info('Trying to download {} to {} (temporary file: {})'.format(url, target, temp.name))
+        download_to_file(url, temp)
+        shutil.copy(temp.name, target)
```
gebner, line 132:
```suggestion
        os.rename(temp.name, target)
```
`shutil.copy` is not atomic.  You'll also need to create the temporary file in the same directory as target ofc.

Vierkantor: Looking into it some more, doing this completely correctly across all OSes is not trivial at all. So I just went with a library instead.

PR #106, opened 2021-08-03
feat(get-cache): Perform a topological walk to find all viable caches

eric-wieser:
Roughly implements what is described in [this zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/get-mathlib-at-commit/near/213010027).
This walks the commit graph, starting from the current commit and going via all parents, until all paths are "blocked" by a commit that has a cache available. The resulting set of caches are unordered with respect to the git graph; none are parents of the other, so all are equally suitable candidates based on the commit graph alone.

For now, we download all suitable caches but then pick one arbitrarily. By printing out all the caches we downloaded, the user can look at their history and make an educated guess at which one will be best for them; for instance by looking at which has a smaller diff with their current commit, or which touches more "foundational" files.

Tested by merging together two random PR branches, which gives the result:
```
$ py -3.8 -m mathlibtools.leanproject -f get-cache
Looking for remote mathlib oleans
Trying to download https://oleanstorage.azureedge.net/mathlib/cbacc5222625f3fb8fb9caa717009cafba045a2d.tar.xz to C:\Users\wiese\.mathlib\cbacc5222625f3fb8fb9caa717009cafba045a2d.tar.xz
No cache available for revision cbacc52226
Looking for remote mathlib oleans
Trying to download https://oleanstorage.azureedge.net/mathlib/dc6adcc85e0197075ea3a2e55c03e854008f5916.tar.xz to C:\Users\wiese\.mathlib\dc6adcc85e0197075ea3a2e55c03e854008f5916.tar.xz
100%|███████████████████████████████████████████████████████████████████| 44.8M/44.8M [00:10<00:00, 4.61MB/s] Found mathlib oleans at https://oleanstorage.azureedge.net/mathlib/
Looking for remote mathlib oleans
Trying to download https://oleanstorage.azureedge.net/mathlib/a2ff4126158907423495941eab9db0995c58a973.tar.xz to C:\Users\wiese\.mathlib\a2ff4126158907423495941eab9db0995c58a973.tar.xz
No cache available for revision a2ff412615
Looking for remote mathlib oleans
Trying to download https://oleanstorage.azureedge.net/mathlib/0323adaf907a25d4bbf67c106ccc520a30395f9f.tar.xz to C:\Users\wiese\.mathlib\0323adaf907a25d4bbf67c106ccc520a30395f9f.tar.xz
100%|███████████████████████████████████████████████████████████████████| 44.8M/44.8M [00:04<00:00, 9.61MB/s] Found mathlib oleans at https://oleanstorage.azureedge.net/mathlib/
No cache was available for cbacc52226.
There are multiple viable caches from parent commits, using the first:
 * dc6adcc85e
 * 0323adaf90
All caches have been downloaded; use `get-cache --rev` to select a different one.
To see the commits in question, run:
  git log --graph cbacc52226 dc6adcc85e^! 0323adaf90^!
```
```
$ git log --graph cbacc52226 dc6adcc85e^! 0323adaf90^!
*   commit cbacc5222625f3fb8fb9caa717009cafba045a2d (HEAD -> eric-wieser/kronecker)
|\  Merge: a2ff412615 dc6adcc85e
| | Author: Eric Wieser <wieser.eric@gmail.com>
| | Date:   Fri Aug 6 10:40:11 2021 +0100
| |
| |     Merge remote-tracking branch 'origin/master' into eric-wieser/kronecker
| |
| * commit dc6adcc85e0197075ea3a2e55c03e854008f5916 (origin/master, origin/HEAD)
|   Author: Yaël Dillies <yael.dillies@gmail.com>
|   Date:   Fri Aug 6 06:59:28 2021 +0000
|
|       feat(order/bounded_lattice): define the `distrib_lattice_bot` typeclass (#8507)
|
|       Typeclass for a distributive lattice with a least element.
|
|       This typeclass is used to generalize `disjoint_sup_left` and similar.
|
|       It inserts itself in the hierarchy between `semilattice_sup_bot, semilattice_inf_bot` and `generalized_boolean_algebra`, `bounded_distrib_lattice`. I am doing it through `extends`.
|
* commit a2ff4126158907423495941eab9db0995c58a973
| Author: Eric Wieser <wieser.eric@gmail.com>
| Date:   Fri Aug 6 10:33:35 2021 +0100
|
|     test
|
* commit 0323adaf907a25d4bbf67c106ccc520a30395f9f (origin/eric-wieser/kronecker)
  Author: faenuccio <filippo.nuccio@univ-st-etienne.fr>
  Date:   Fri Aug 6 11:09:54 2021 +0200

      modified docs
```

This also removes the retry logic, which shouldn't need to be special cased now that we use `atomicwrites` to never leave a partial cache. If we want it back, it probably belongs in `get_mathlib_archive`.

You can try this out with `python3 -m pipx install git+https://github.com/eric-wieser/mathlib-tools@get-cache-search`

### DIFF
--- a/mathlibtools/git_helpers.py
+++ b/mathlibtools/git_helpers.py
@@ -1,4 +1,4 @@
-import git
+import git   # type: ignore
 from typing import Callable, Iterator, Tuple, List
 
 
--- a/mathlibtools/git_helpers.py
+++ b/mathlibtools/git_helpers.py
@@ -2,6 +2,11 @@ import git   # type: ignore
 from typing import Callable, Iterator, Tuple, List
 
 
+def short_sha(rev: git.Commit) -> str:
+    """ Truncate `rev.hexsha` without ambiguity """
+    return rev.repo.git.rev_parse(rev.hexsha, short=True)
+
+
 def visit_ancestors(rev: git.Commit) -> Iterator[Tuple[git.Commit, Callable]]:
     r"""
     Iterate over the ancestors of the commit `rev` in topological order, with
--- a/mathlibtools/git_helpers.py
+++ b/mathlibtools/git_helpers.py
@@ -1,7 +1,5 @@
 import git
-from typing import Callable, TypeVar, Iterator, Tuple
-
-T = TypeVar('T')
+from typing import Callable, Iterator, Tuple, List
 
 
 def visit_ancestors(rev: git.Commit) -> Iterator[Tuple[git.Commit, Callable]]:
@@ -20,12 +18,7 @@ def visit_ancestors(rev: git.Commit) -> Iterator[Tuple[git.Commit, Callable]]:
     where ``A`` is the root commit and ``K`` and ``L`` are tips of branches.
     The following code runs against this commit graph
 
-    >>> def filter_fun(c):
-    ...     print('visited', c)
-    ...     if c in {B, F, G}:
-    ...         return c
     >>> for c, prune in visit_ancestors(L):
-    ...     print('visited', c)
     ...     if c in {B, F, G}:
     ...         prune()
     ...         print('found  ', c)
@@ -33,18 +26,18 @@ def visit_ancestors(rev: git.Commit) -> Iterator[Tuple[git.Commit, Callable]]:
     ...         print('visited', c)
     visited L
     visited J
-    visited I
-    visited E
-    found   F
     visited H
+    visited I
+    found   F
     found   G
+    visited E
 
     The exact order these commits appear in depends on the order of parents in
     merge commits, but independent of this ``B`` will never be visited as it is
     a parent of ``F`` and ``G``, and the sort order is topological.
     """
     repo = rev.repo
-    pruned_commits = []  # the commits to ignore along with their ancestors
+    pruned_commits : List[git.Commit] = []  # the commits to ignore along with their ancestors
     skip_n = 0  # the index to resume the iteration
     while True:
         args = [rev] + ['--not'] + pruned_commits
--- a/mathlibtools/git_helpers.py
+++ b/mathlibtools/git_helpers.py
@@ -9,10 +9,21 @@ def short_sha(rev: git.Commit) -> str:
 
 def visit_ancestors(rev: git.Commit) -> Iterator[Tuple[git.Commit, Callable]]:
     r"""
-    Iterate over the ancestors of the commit `rev` in topological order, with
-    the option to prune parents of visited commits.
+    Iterate over history, optionally pruning all ancestors of a given commit.
 
-    Consider the commit graph::
+    This iterates backwards over history starting at `rev` and traversing the
+    commit graph in topological (as opposed to date) order, ensuring that child
+    commits are always visited before any of their parent commits. In this
+    sense, this function is like ``repo.iter_commits(rev, topo_order=True)``.
+
+    The key difference from ``iter_commits`` is that this version yields
+    ``commit, prune`` pairs, where ``prune`` is a function accepting no
+    arguments. If ``prune()`` is called, then the iterator will not visit any
+    of the commits which are ancestors of ``commit``; that is, the history
+    "tree" from that point backwards is pruned.
+
+    As an example, consider a repository with the commit graph below, where
+    ``A`` is the root commit and ``K`` and ``L`` are tips of branches::
 
         A -- B -- E -- I -- J -- L
               \       /    /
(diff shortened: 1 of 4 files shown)


gebner:
I believe part of the problem is that the gitpython package only got typing annotations in version 3.1.19, but 3.1.19 and 3.1.20 were yanked from pypi.

eric-wieser:
Yes, it looks like the remaining mypy errors aren't something that I can fix easily.

eric-wieser:
Ran into a gitpython bug while writing the test: https://github.com/gitpython-developers/GitPython/pull/1314. CI is green now.

mathlibtools/git_helpers.py
```diff
@@ -0,0 +1,65 @@
...
+
+def short_sha(rev: git.Commit) -> str:
+    """ Truncate `rev.hexsha` without ambiguity """
+    return rev.repo.git.rev_parse(rev.hexsha, short=True)
+
+
+def visit_ancestors(rev: git.Commit) -> Iterator[Tuple[git.Commit, Callable]]:
+    r"""
+    Iterate over the ancestors of the commit `rev` in topological order, with
+    the option to prune parents of visited commits.
```
PatrickMassot, line 13: I think this is the third time I come here to start reviewing this PR and I stop at the inscrutable docstring. I can vaguely understand what kind of thing you mean by "topological order" (although this is far from explicit) but then I have no idea what the second part means. Reading the example helps to guess, but you should really explain it properly.

PatrickMassot (commented):
I really don't think we should download all available oleans. Of course this will be an extremely rare event, but it really feels wrong. I think we should check whether the file exist and ask users to choose.

eric-wieser:
> I really don't think we should download all available oleans. Of course this will be an extremely rare event, but it really feels wrong. I think we should check whether the file exist and ask users to choose.

This event is very common when using gitpod to review PRs, and it's nice to have gitpod load up with all the caches downloaded.

I'm happy to hide this behavior behind a flag though - do you have a preference? Maybe `get-cache --download=all` vs `download=ask`?

mathlibtools/git_helpers.py
eric-wieser: Thanks for taking a look at it!

Perhaps a reference to the git documentation would be helpful here regarding the meaning I intend by "topological". I'll revise this docstring. Does the example further down the docstring help explain the behavior? I'm happy to spell things out more if you can point out where explanation is most needed.

eric-wieser: I've rewritten the docstring from scratch - let me know if it's better.

PatrickMassot: This is much better, thanks!

PatrickMassot:
> > I really don't think we should download all available oleans. Of course this will be an extremely rare event, but it really feels wrong. I think we should check whether the file exist and ask users to choose.
> 
> This event is very common when using gitpod to review PRs, and it's nice to have gitpod load up with all the caches downloaded.
> 
> I'm happy to hide this behavior behind a flag though - do you have a preference? Maybe `get-cache --download=all` vs `download=ask` and `download=exact` (the current behavior)?

Oh, I see. I guess we can easily configure gitpod to fetch everything and ask users by default. So I vote for asking by default and having a flag to download all.

eric-wieser:
Having thought about it a bit more, I think the only behaviors worth having are:
* Download exact match only, do not attempt a history search (for when Kevin is "in a field" and on a metered connection)
* Download an exact match, else search for but do not download all available ancestors. If the exact match is not available, error with a message telling the user which commit to ask for instead. The rationale here is that the user probably does not know how to choose on the spot, and will likely have to run a git command to help decide which cache to choose.
* Download the first match. This is what mathlib CI would use
* Download all matches. This is a version of 2 for people who just want to download everything now, and decide which one they needed later.

Does that list look sensible to you?

PatrickMassot:
Sounds good.

eric-wieser:
Ok, updated. Here's the new behavior:
```
```
$ py -3.8 -m mathlibtools.leanproject get-cache --fallback=none
Looking for mathlib oleans for 7b95062ac8
  locally...
  remotely...
Failed to fetch cached oleans
```
```
$ py -3.8 -m mathlibtools.leanproject get-cache --fallback=show
Looking for mathlib oleans for 7b95062ac8
  locally...
  remotely...
No cache available for revision 7b95062ac8
Looking for mathlib oleans for 580900f225
  locally...
  Found local mathlib oleans
Looking for mathlib oleans for 5e59fb48eb
  locally...
  Found local mathlib oleans
No cache was available for 7b95062ac8.
There are multiple viable caches from parent commits:
 * 580900f225
 * 5e59fb48eb
To see the commits in question, run:
  git log --graph 7b95062ac8 580900f225^! 5e59fb48eb^!
Run `leanproject get-cache --rev` on one of the available commits above.
Failed to fetch cached oleans
```
```
$ py -3.8 -m mathlibtools.leanproject get-cache --fallback=download-first
Looking for mathlib oleans for 7b95062ac8
  locally...
  remotely...
No cache available for revision 7b95062ac8
Looking for mathlib oleans for 580900f225
  locally...
  Found local mathlib oleans
Looking for mathlib oleans for 5e59fb48eb
  locally...
  Found local mathlib oleans
No cache was available for 7b95062ac8.
There are multiple viable caches from parent commits:
 * 580900f225
 * 5e59fb48eb
To see the commits in question, run:
  git log --graph 7b95062ac8 580900f225^! 5e59fb48eb^!
Downloading first cache
Applying cache
```
(with `-f` to force a download this time)
```
$ py -3.8 -m mathlibtools.leanproject -f get-cache --fallback=download-all
Looking for mathlib oleans for 7b95062ac8
  remotely...
No cache available for revision 7b95062ac8
Looking for mathlib oleans for 580900f225
  remotely...
  Found remote mathlib oleans
Looking for mathlib oleans for 5e59fb48eb
  remotely...
  Found remote mathlib oleans
No cache was available for 7b95062ac8.
There are multiple viable caches from parent commits:
 * 580900f225
 * 5e59fb48eb
To see the commits in question, run:
  git log --graph 7b95062ac8 580900f225^! 5e59fb48eb^!
Downloading all caches
  5e59fb48eb: 100%|█████████████████████████████████████████████| 45.1M/45.1M [00:06<00:00, 6.92MB/s]
  580900f225: 100%|█████████████████████████████████████████████| 44.0M/44.0M [00:08<00:00, 5.45MB/s]
  Applying cache
```

**Comment by PatrickMassot** [comment]
Could you please add docstrings to all those new classes?

**Comment by eric-wieser** [comment]
Are the new docstrings enough, or do you want more?

**Comment by PatrickMassot** [comment]
When there is a local cache for the desired commit, the message is:
```
Looking for mathlib oleans for 7b5c60db3
  locally...
  Found local mathlib oleans
Downloading matching cache
Applying cache
```
and the "Downloading matching cache" is weird. Is it only a misleading message or is it going down the wrong route?

eric-wieser:
It's just a misleading message, there'd be a progress bar if a download occurred.

PatrickMassot:
That's what I thought. Could you fix that please?

eric-wieser:
I've adjusted the wording

eric-wieser:
Would you mind taking another look at this, @PatrickMassot?

eric-wieser:
Thanks!

PatrickMassot:
Thanks a lot for this very nice contribution!

alreadydone:
I was getting error `type object 'tqdm' has no attribute 'wrapattr'` after recently upgrading mathlib-tools from 0.0.9 to 1.1.0. Upgrading tqdm solves this problem; turns out this is compatible with latest tqdm 4.62.3 but not with old tqdm 4.26.0. However now I'm getting `'WindowsPath' object has no attribute 'decode'` after the cache has been downloaded ...
It seems when pip upgrades mathlib-tools it doesn't install the exact versions specified in `pyinstaller-requirements.txt`, e.g.
`Requirement already satisfied, skipping upgrade: atomicwrites in e:\users\xjy\anaconda3\lib\site-packages (from mathlibtools) (1.2.1)`
but the required version is 1.4.0.

PR #108, opened 2021-08-07
Recommend homebrew for macOS in the README.

Julian:


FrederickGeek8:
Just FYI I had to do `brew install lean` after `brew install mathlibtools` on macOS. Not sure if it's worth raising an issue over though.

Julian:
Hi! That isn't recommended, the lean community prefers using elan for managing versions, which `brew install elan` will get you. The full [install script](https://leanprover-community.github.io/install/macos.html) which is referenced in the README already I think should do so for you as well. -- can you tell me more about what else you'd ran besides the command from this PR?

FrederickGeek8:
Ah! Good point. Sorry, I'm new here.

Perhaps it would make sense to also edit [leanprover-community/leanprover-community.github.io/blob/newsite/templates/install/macos_details.md](https://github.com/leanprover-community/leanprover-community.github.io/blob/newsite/templates/install/macos_details.md) in light of these changes. It seems like this guidance diverges from the script which itself diverges from the "detailed" script page.

Julian:
No worries, that is indeed a very good idea, will do! I knew there was a page I'd forgotten somewhere...

PatrickMassot:
Just in case someone is waiting for me, let me repeat that I have no access to user-hostile operating systems, so I won't comment on any MacOS-specific PR. But I'm happy to merge such PRs once at least one more MacOS user approves.

Julian:
@FrederickGeek8 do you agree with this changeset given we've changed the other one too now? If so probably your lgtm is good enough to satisfy Patrick :)

FrederickGeek8:
Looks good to me :)

PR #112, opened 2021-08-17
feat(get-cache): Show a vague indicator of progress when untar-ing the cache

eric-wieser:
This at least prints out the number of files extracted so far, which clues the user into the fact that it is making progress.
For instance, on my machine this now shows:
```
Applying cache
  files extracted: 1651 [00:18, 90.61/s]
```
where the second line is new.
Instrumenting the extraction doesn't appear to impact performance in any meaningful way.

I tried to make it show a proper progress bar, but that requires knowing the contents of the tar file in advance.
It doesn't seem to be safe to iterate the members of a `tarfile.TarFile` from one thread and read them from another, so I didn't bother exploring further.

### DIFF
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -110,7 +110,8 @@ def pack(root: Path, srcs: Iterable[Path], target: Path) -> None:
 def unpack_archive(fname: Union[str, Path], tgt_dir: Union[str, Path]) -> None:
     """ Alternative to `shutil.unpack_archive` that shows progress"""
     with tarfile.open(fname) as tarobj:
-        tarobj.extractall(str(tgt_dir), members=tqdm(tarobj))
+        tarobj.extractall(
+            str(tgt_dir), members=tqdm(tarobj, desc='  files extracted', unit=''))
 
 class OleanCache:
     """ A reference to a cache of oleans for a single commit.
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -14,6 +14,7 @@ import contextlib
 import enum
 from datetime import datetime
 import concurrent.futures
+import tarfile
 from typing import Iterable, IO, Union, List, Tuple, Optional, Dict, TYPE_CHECKING
 from tempfile import TemporaryDirectory
 import shutil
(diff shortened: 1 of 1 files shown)


PatrickMassot:
I tried to use it but I have no idea what it means, so I don't think it's useful in its current form.

eric-wieser:
Is this better?
```
Applying cache
  files extracted: 1651 [00:18, 90.61/s]
```

PatrickMassot:
Is there any way to get the number of files in the archive without uncompressing it?

eric-wieser:
I couldn't find a way, I don't think tar files contain that information. Enumerating the files takes about as long as extracting them.

jcommelin:
We could save that info somewhere (eg, a separate file `longhash.meta.txt`)

eric-wieser:
I think the real problem here is that `tarfile` is slow compared to the Unix `tar` command - I'm sure `tar` could count the entries very quickly, but python takes around 10s for me.

PatrickMassot:
We could store that count on Azure as well, but maybe that's going too far for a feature that nobody ever requested...

jcommelin:
> I think the real problem here is that `tarfile` is slow compared to the Unix `tar` command - I'm sure `tar` could count the entries very quickly, but python takes around 10s for me.

This seems like something that a benchmark can quantify.

gebner:
Even on my machine it takes 2 seconds to list the contents of the tar archive (with the command-line tar tool).  We can't count the files more quickly than that.

eric-wieser:
If I change the implementation to:

```
    with tarfile.open(fname) as tarobj:
        members = list(tqdm(tarobj, desc='  files found', unit=''))
        tarobj.extractall(
            str(tgt_dir), members=tqdm(members, desc='  files extracted', unit='', total=len(members)))
```
I get the nicer output
```
Applying cache
  files found: 1652 [00:15, 106.60/s]
  files extracted: 100%|████████████████████████████████████████████| 1652/1652 [00:22<00:00, 73.25/s]
```
but this takes ~1.75 times longer so doesn't seem worth it to me.

> but maybe that's going too far for a feature that nobody ever requested...

I kind of feel the same way - my goal in this PR was simply to reassure the user that progress is being made, because it often hangs for quite some time after the last log message. Showing the total amount of work left would be great, but doing so without inflating the amount of time it takes in the first place seems hard.

gebner:
Can we figure out how much of the *compressed* file has been read yet?  This could be a reasonable (and fast to compute) proxy for the unpacking progress.

eric-wieser:
I thought about but didn't try that; the outline would be:
* Open the tarfile in streaming mode to prevent random file access
* Either:
  * Somehow get the *uncompressed* file size, and use `tqdm.wrapattr` on `tarfile.fileobj` to measure progress of uncompressed data
  * Somehow get the *compressed* file handle which is easy to measure the size of, and use `tqdm.wrapattr` on that to mark the progress

jcommelin:
I do like the idea of reassuring the user that something is happening.

eric-wieser:
I don't plan to spend much more time looking at tar files; can we put this (or a tweaked version regarding log messages) in, and leave the "efficiently compute the total size" problem to a future PR? I'd argue a progress indicator without a total is still an improvement over no indicator at all.

PR #113, opened 2021-08-22
fix: various improvements to get-cache

eric-wieser:
Apologies for the size of this PR; my impression was that the semantics of some of the commands were mixed up, and it was easier to change them all at once

### Before this PR

* `leanproject mk-cache --force` on a dirty tree:
  * Creates a cache entry for `HEAD` that represents the dirty state of the working tree (which is not HEAD!)

* `leanproject get-cache`
  * on Mathlib:
    * Does a full history search
    * Unpacks the full archive over the working tree. If this archive comes from azure, it contains only oleans; if it comes from `mk-cache` it contains all the lean source files too.
  * on another project:
    * Uses only an exact match
    * Always overwrites local lean files, meaning that `get-cache --rev old_revision` acts much like `git checkout old_revision -- .`

* `leanproject get-mathlib-cache`
  * on mathlib:
    * Behaved identically to `leanproject get-cache`
  * on another project:
    * Does a full history search
    * Checks out lean files by running `git reset` on mathlib
 
### After this PR

* `leanproject mk-cache --force` on a dirty tree:
  * is not supported. `--force` is now only for overwriting the cache if it exists

* `leanproject get-cache` on any project:
  * Does a full history search
  * Unpacks only olean files
  * No longer supports `--force`. It only made sense in the presence of the destructive behavior regarding .lean files, which is no longer present.

* `leanproject get-mathlib-cache`
  * No longer supports `--rev` 
  * on mathlib:
    * Behaved identically to `leanproject get-cache`, spits out a warning telling the user to use `get-cache` directly.
  * on another project:
    * Does a full history search
    * Checks out mathlib lean files by running `git reset` on mathlib, then applying lean files from the tarball.
 
This replaces #96

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## 1.1.0dev (unreleased)
 
 * `get-mathlib-cache` no longer understands `--rev`; if you want to use a
-  different mathlib version, edit your `leanproject.toml`.
+  different mathlib version, edit your `leanproject.toml`. If you are trying to get
+  the cache when working on mathlib itself, use `get-cache`.
 * Add `--fallback` to `get-cache` for traversing the git history to find an
   approximate cache.
 * `get-cache` no longer modifies `lean` files in the working directory.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change log
 
+## 1.1.0dev (unreleased)
+
+* `get-mathlib-cache` no longer understands `--rev`; if you want to use a
+  different mathlib version, edit your `leanproject.toml`.
+* Add `--fallback` to `get-cache` for traversing the git history to find an
+  approximate cache.
+* `get-cache` no longer modifies `lean` files in the working directory.
+* `mk-cache --force` no longer associates the cache with the current HEAD
+  revision, but instead creates a temporary commit in order to obtain a SHA1 id.
+
 ## 1.0.0 (2020-11-10)
 
 * Only look for .xz archives
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,11 @@
 
 * `get-mathlib-cache` no longer understands `--rev`; if you want to use a
   different mathlib version, edit your `leanproject.toml`. If you are trying to get
-  the cache when working on mathlib itself, use `get-cache`.
+  the cache when working on mathlib itself, use `get-cache --rev`.
 * Add `--fallback` to `get-cache` for traversing the git history to find an
   approximate cache.
-* `get-cache` no longer modifies `lean` files in the working directory.
+* `get-cache` no longer modifies `.lean` files in the working directory.
 * `mk-cache --force` no longer permits the working tree to be dirty.
-  revision, but instead creates a temporary commit in order to obtain a SHA1 id.
 
 ## 1.0.0 (2020-11-10)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Change log
 
+## 1.1.0dev (unreleased)
+
+* `get-mathlib-cache` no longer understands `--rev`; if you want to use a
+  different mathlib version, edit your `leanproject.toml`.
+* Add `--fallback` to `get-cache` for traversing the git history to find an
+  approximate cache.
+* `get-cache` no longer modifies `lean` files in the working directory.
+* `mk-cache --force` no longer associates the cache with the current HEAD
```
jcommelin, line 10: Personally, I think I would prefer if `leanproject` would categorically deny working on a dirty repo. It will probably save people quite some stress and tears.

eric-wieser: This happens to me all the time:
* Checkout a branch
* Start editing a file `foo.lean`, and save it
* Remember I didn't run `get-cache` yet

With this PR, `get-cache --force` will get the fresh oleans from azure, without causing the tears it currently causes by throwing away my changes to `foo.lean`.

jcommelin: well, it creates a `git commit`, which might confuse users that are not very gitperienced.
What does `leanproject` do without the `--force`? Will it tell me that the directory is dirty, and that I should fix that or use `--force`? That would be good, I think.

jcommelin: I personally like having the choice of doing `git stash` instead of `git commit`. So if `leanproject` doesn't do any git-stuff unless I `--force` it, then I'm happy.

eric-wieser: It doesn't really run either `git commit` or `git stash`; the commit it creates is invisible to any novice user of git, and only there if you're really looking for it.

At any rate, this comment chain might be more suited for #115

jcommelin: Understood.

eric-wieser:
```suggestion
* `mk-cache --force` no longer permits the working tree to be dirty.
```

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Change log
 
+## 1.1.0dev (unreleased)
+
+* `get-mathlib-cache` no longer understands `--rev`; if you want to use a
+  different mathlib version, edit your `leanproject.toml`.
```
eric-wieser, line 6:
```suggestion
  different mathlib version, edit your `leanproject.toml`. If you are trying to get
  the cache when working on mathlib itself, use `get-cache`.
```

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+## 1.1.0dev (unreleased)
+
+* `get-mathlib-cache` no longer understands `--rev`; if you want to use a
+  different mathlib version, edit your `leanproject.toml`. If you are trying to get
+  the cache when working on mathlib itself, use `get-cache`.
+* Add `--fallback` to `get-cache` for traversing the git history to find an
+  approximate cache.
+* `get-cache` no longer modifies `lean` files in the working directory.
+* `mk-cache --force` no longer permits the working tree to be dirty.
+  revision, but instead creates a temporary commit in order to obtain a SHA1 id.
```
jcommelin, line 12: This is no longer accurate, is it?

eric-wieser:
```suggestion
* `mk-cache --force` no longer permits the working tree to be dirty.
```

PR #114, opened 2021-08-31
wrap filenames containing hyphens in french quotes when making all.lean

alexjbest:
This caused problems for list_decls and other tools, as in https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there.20code.20for.20X.3F/topic/Create.20a.20blueprint/near/251398961

Probably users should still not use hyphens in filenames.

### DIFF
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -846,7 +846,7 @@ class LeanProject:
 
         # helper function to wrap file paths in double quotes if they
         # contain dashes, so that lean accepts them as imports
-        def make_safe(s):
+        def escape_identifier(s : str) -> str:
             if re.match(LEAN_UNESCAPED_IDENTIFIER_RE, s):
                 return s
             return "«" + s + "»"
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -855,10 +855,10 @@ class LeanProject:
 
         with (self.src_directory/'all.lean').open('w') as all_file:
             for path in self.src_directory.glob('**/*.lean'):
-                rel = str(path.relative_to(self.src_directory).with_suffix(''))
-                if rel == 'all':
+                rel = path.relative_to(self.src_directory).with_suffix('')
+                if rel == Path('all'):
                     continue
-                all_file.write('import ' + ".".join(map(escape_identifier, rel.split(os.path.sep))) + '\n')
+                all_file.write('import ' + ".".join(map(escape_identifier, rel.parts)) + '\n')
 
     def list_decls(self) -> Dict[str, DeclInfo]:
         """Collect declarations seen from this project, as a dictionary of
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -114,6 +114,14 @@ def unpack_archive(fname: Union[str, Path], tgt_dir: Union[str, Path]) -> None:
         tarobj.extractall(
             str(tgt_dir), members=tqdm(tarobj, desc='  files extracted', unit=''))
 
+def escape_identifier(s : str) -> str:
+    """ Helper function to wrap identifiers in double french quotes if they
+    need to be wrapped by lean, we use this for file paths so that lean accepts
+    them as imports"""
+    if re.fullmatch(LEAN_UNESCAPED_IDENTIFIER_RE, s):
+        return s
+    return "«" + s + "»"
+
 class OleanCache:
     """ A reference to a cache of oleans for a single commit.
 
@@ -844,19 +852,12 @@ class LeanProject:
     def make_all(self) -> None:
         """Creates all.lean importing everything from the project"""
 
-        # helper function to wrap file paths in double quotes if they
-        # contain dashes, so that lean accepts them as imports
-        def escape_identifier(s : str) -> str:
-            if re.fullmatch(LEAN_UNESCAPED_IDENTIFIER_RE, s):
-                return s
-            return "«" + s + "»"
-
         with (self.src_directory/'all.lean').open('w') as all_file:
             for path in self.src_directory.glob('**/*.lean'):
                 rel = str(path.relative_to(self.src_directory).with_suffix(''))
                 if rel == 'all':
                     continue
-                all_file.write('import ' + ".".join(map(make_safe, rel.split(os.path.sep))) + '\n')
+                all_file.write('import ' + ".".join(map(escape_identifier, rel.split(os.path.sep))) + '\n')
 
     def list_decls(self) -> Dict[str, DeclInfo]:
         """Collect declarations seen from this project, as a dictionary of
--- a/mathlibtools/lib.py
+++ b/mathlibtools/lib.py
@@ -70,7 +70,11 @@ DOWNLOAD_URL_FILE = DOT_MATHLIB/'url'
 
 MATHLIB_URL = 'https://github.com/leanprover-community/mathlib.git'
 LEAN_VERSION_RE = re.compile(r'(.*)\t.*refs/heads/lean-(.*)')
-LEAN_UNESCAPED_IDENTIFIER_RE = re.compile(r"""(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟](?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ])*""")
+# This regex is from [1] and implements the logic at [2].
+# [1]: https://github.com/leanprover/vscode-lean/blob/2b43982c4c6305a0f20f156152a60613a6f1a683/syntaxes/lean.json#L193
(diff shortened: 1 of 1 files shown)


mathlibtools/lib.py
```diff
@@ -842,12 +842,20 @@ def reduce_imports_sed(self, file: str) -> Iterable[str]:
 
     def make_all(self) -> None:
         """Creates all.lean importing everything from the project"""
+
+        # helper function to wrap file paths in double quotes if they
+        # contain dashes, so that lean accepts them as imports
+        def make_safe(s):
+            if "-" in s:
+                return "«" + s + "»"
+            return s
```
eric-wieser, line 851: Does it make sense to use the regex from here?

https://github.com/leanprover/vscode-lean/blob/2b43982c4c6305a0f20f156152a60613a6f1a683/syntaxes/lean.json#L184-L186

alexjbest: I'm not sure what the rules are for imports, and even a regex for the valid names wouldn't tell us which ones need escaping. I considered adding `.` here also, but I don't know if anyone would even try to use a `.` in a lean file path?

eric-wieser: It looks like import tokens are no different to regular identifiers:

https://github.com/leanprover-community/lean/blob/4f0963ee6cc8380b0178fa9930797f0cfb730694/src/frontends/lean/parser.cpp#L2682

https://github.com/leanprover-community/lean/blob/4f0963ee6cc8380b0178fa9930797f0cfb730694/src/frontends/lean/parser.cpp#L2711-L2712

Which suggests the above regex is the right one.

eric-wieser: For reference, this is how lean itself decides when to escape identifiers:
```
https://github.com/leanprover-community/lean/blob/65ad4ffdb3abac75be748554e3cbe990fb1c6500/src/util/name.cpp#L65-L83
```

alexjbest: Oh nice detective work! So the regex you posted is the right one! Wow I'm surprised that so much stuff doensn't need escaping I guess

mathlibtools/lib.py
```diff
@@ -70,6 +70,7 @@ class LeanProjectError(Exception):
 
 MATHLIB_URL = 'https://github.com/leanprover-community/mathlib.git'
 LEAN_VERSION_RE = re.compile(r'(.*)\t.*refs/heads/lean-(.*)')
+LEAN_UNESCAPED_IDENTIFIER_RE = re.compile(r"""^(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟](?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ])*(\\.(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟](?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ])*)*$""")
```
eric-wieser, line 73: I don't think you want to match dotted names here:
```suggestion
LEAN_UNESCAPED_IDENTIFIER_RE = re.compile(r"""^(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟](?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ])*$""")
```

mathlibtools/lib.py
```diff
@@ -842,12 +843,20 @@ def reduce_imports_sed(self, file: str) -> Iterable[str]:
 
     def make_all(self) -> None:
         """Creates all.lean importing everything from the project"""
+
+        # helper function to wrap file paths in double quotes if they
+        # contain dashes, so that lean accepts them as imports
+        def make_safe(s):
+            if re.match(LEAN_UNESCAPED_IDENTIFIER_RE, s):
```
eric-wieser, line 850: Note that `match` matches at the start of the string, making the `^` in the regex redundant. If you use `fullmatch` instead, then the `$` is redundant too.

mathlibtools/lib.py
alexjbest: I tested the dotted names i.e `a.a` using the original regex at https://pythex.org/ and they don't match, which is the right behaviour right? Such filenames would need escaping?

mathlibtools/lib.py
alexjbest: cool thanks

mathlibtools/lib.py
eric-wieser: You've incorrectly copied the `\\.`, which was originally in a non-escaped string and represented the regex `\.` aka a single `.`. Here it means "a `\` followed by a single character", which is not what you meant.

mathlibtools/lib.py
```diff
@@ -70,6 +70,7 @@ class LeanProjectError(Exception):
 
 MATHLIB_URL = 'https://github.com/leanprover-community/mathlib.git'
 LEAN_VERSION_RE = re.compile(r'(.*)\t.*refs/heads/lean-(.*)')
+LEAN_UNESCAPED_IDENTIFIER_RE = re.compile(r"""(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟](?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ])*(\\.(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟](?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ])*)*""")
```
eric-wieser, line 73: Repeating the previous suggestionsince the diff moved:
```suggestion
LEAN_UNESCAPED_IDENTIFIER_RE = re.compile(r"""(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟](?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ])*""")
```

alexjbest: cool

mathlibtools/lib.py
```diff
@@ -70,6 +70,7 @@ class LeanProjectError(Exception):
 
 MATHLIB_URL = 'https://github.com/leanprover-community/mathlib.git'
 LEAN_VERSION_RE = re.compile(r'(.*)\t.*refs/heads/lean-(.*)')
+LEAN_UNESCAPED_IDENTIFIER_RE = re.compile(r"""(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟](?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ])*""")
```
eric-wieser, line 73: Let's leave a breadcrumb here:
```suggestion
# This regex is from [1] and implements the logic at [2].
# [1]: https://github.com/leanprover/vscode-lean/blob/2b43982c4c6305a0f20f156152a60613a6f1a683/syntaxes/lean.json#L193
# [2]: https://github.com/leanprover-community/lean/blob/65ad4ffdb3abac75be748554e3cbe990fb1c6500/src/util/name.cpp#L65-L83
LEAN_UNESCAPED_IDENTIFIER_RE = re.compile(
  r"(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟](?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ])*")
```
Triple quotes aren't needed here either, since there's no line break inside.

mathlibtools/lib.py
```diff
@@ -113,6 +118,15 @@ def unpack_archive(fname: Union[str, Path], tgt_dir: Union[str, Path]) -> None:
         tarobj.extractall(
             str(tgt_dir), members=tqdm(tarobj, desc='  files extracted', unit=''))
 
+def escape_identifier(s : str) -> str:
+    """ Helper function to wrap _pieces_ of identifiers in double french quotes
+    if they need to be wrapped by lean, we use this for file paths so we also escape
+    strings of the form `a.a` even though they are in valid identifiers.
```
alexjbest, line 124:
```suggestion
    strings of the form `a.a` even though they are valid identifiers.
```

eric-wieser (approved):

PR #115, opened 2021-09-01
feat(leanproject mk-cache --force): create a temporary commit to house dirty caches

eric-wieser:
The cache is indexed by the SHA1 of the working tree that produced it.
Previously `mk-cache --force` ignored this indexing scheme, replacing the cache for SHA1 with a non-matching cache.

The new behavior is to create a temporary commit containing the changes in the working tree, and associate the cache with that commit.
The commit is not created on the users current branch, so does not interfere with the user's use of git.

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/What.20is.20.60leanproject.20mk-cache.20--force.60.20for/near/250270614)

### DIFF
--- a/mathlibtools/leanproject.py
+++ b/mathlibtools/leanproject.py
@@ -196,7 +196,10 @@ def get_project(name: str, new_branch: bool, directory: str = '') -> None:
 @click.option('--force', default=False, is_flag=True,
               help='Make cache even if the repository is dirty or cache exists.')
 def mk_cache(force: bool = False) -> None:
-    """Cache olean files."""
+    """Cache olean files.
+
+    If run with `--force` on a dirty repository, this creates a temporary commit
+    to associate the dirty cache with."""
     try:
         proj().mk_cache(force)
     except LeanDirtyRepo as err:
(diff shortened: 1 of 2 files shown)


jcommelin:
Hmm, I still think that users could get mightily confused by this behaviour.
```shell
# In a dirty working tree on branch foobar
leanproject mk-cache --force
git checkout master
leanproject get-cache
# do stuff
git checkout foobar
leanproject get-cache # fail because the sha of this revision doesn't match the temporary commit, so no caches were made for this revision
```
So users have to be very conscious that the should save that sha that is posted back to them. If you run this at 22:37 before you go to bed, and close your terminal, the sha is lost, and only `git ref-log` magic can retrieve it for you.

eric-wieser:
I agree; the main argument I have for this behavior is it's less harmful than the current behavior. We could also just remove `mk-cache --force` entirely.

Regarding `reflog` magic; we could have a special `leancache` "ref" (but not branch) such that `git reflog leancache` shows the history of applied caches.

jcommelin:
I vote in favour of removing `--force`. If the directory is dirty, tell the user to clean up the mess. That's the most predictable behaviour.

PR #116, opened 2021-09-17
feat(mathlibtools): expose mathlibtools.__version__

eric-wieser:
This allows someone using the python part of the API to get the version number.

This also pre-emptively solves any problems that unicode characters in the readme would cause.

### DIFF
--- a/mathlibtools/__init__.py
+++ b/mathlibtools/__init__.py
@@ -1,0 +1,1 @@
+from ._version import __version__
--- a/mathlibtools/_version.py
+++ b/mathlibtools/_version.py
@@ -1,0 +1,8 @@
+# Package versioning solution originally found here:
+# http://stackoverflow.com/q/458550
+
+# Store the version here so:
+# 1) we don't load dependencies by storing it in __init__.py
+# 2) we can import it in setup.py for the same reason
+# 3) we can import it into your module
+__version__ = '1.0.0'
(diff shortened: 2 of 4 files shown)


Julian:
FWIW in modern packaging this technically isn't necessary, the "official" way to do this for any package is just `importlib.metadata.version("mathlibtools")`.

eric-wieser:
Sure, but that requires a version bump to at least Python 3.8, from:

https://github.com/leanprover-community/mathlib-tools/blob/b6e31f31e035b22a954671416ab46baccc293573/setup.py#L30

PatrickMassot:
I would prefer to continue supporting all python version supported by the python foundation (see https://devguide.python.org/#status-of-python-branches).

Julian:
This PR was merged already, I don't know why I mentioned it -- but on older Python versions it works as well, there's a [backport available](https://pypi.org/project/importlib-metadata/).

eric-wieser:
@Julian, do you have a reference for `__version__` actually being recommended against?

Julian:
https://www.python.org/dev/peps/pep-0396/#pep-rejection has a small note which you perhaps could take as discouragement.

eric-wieser:
Ah, here's the mailing list thread: https://mail.python.org/archives/list/python-dev@python.org/thread/EY44EMZ2R4RUMUUIVEFFVOXM2H3L427Z/#4XWIWIKDJJRWVKNGIOENJ2ZJVQR4PP23

At any rate I don't think any action is needed on mathlib-tools, but I'm interested in the context of other python packages.

PR #121, opened 2022-01-05
Remove use of deprecated logging methods.

Julian:
Logger.warn is spelled Logger.warning.


PatrickMassot:
Do you have references about this deprecation? Which python versions use the new spelling?

Julian:
Sorry, I was lazy -- `logger.warn` was deprecated for 10 years or so at least, it's from before the module was incorporated into the standard library:

```
⊙  python3 -W default -c 'import logging; logging.getLogger().warn("Hello")'                                                                                                        julian@Airm
<string>:1: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
Hello
```

See e.g. https://bugs.python.org/issue13235 -- which looks like it says it was deprecated at least back since 3.3.

PatrickMassot:
Thanks

Issue #120, opened 2021-11-01
'leanproject new' fails with lean 4 installed


avigad:
If one has lean 3 and lean 4 both installed with lean 4 as the system default, the behavior of `leanproject new test` is cryptic. 

What happens is that `leanproject` calls `leanpkg new test`, but `elan` interprets that as a call to the lean 4 `leanpkg`, which does not have a `new` command.

PatrickMassot:
This has been reported on Zulip as well, but I don't quite know what to do, except hoping that people who have Lean 4 as their default Lean know what they are doing.

PatrickMassot:
Thanks for you contribution. We are closing all pull requests and issues since this tool no longer has any relevance in the Lean 4 era. This repository will now be archived.

PR #122, opened 2022-01-05
Give a better error from leanproject new when elan default is set to Lean 4.

Julian:
Closes: #120

Perhaps this is enough of a hint.


PatrickMassot:
If you have time to work on this then I think it would be even better to allow passing an elan toolchain name to `leanproject new`. Even better would be to ask the user whether they want to use the latest available lean3 toolchain.

Julian:
Ok, probably not in the next few days but maybe next week I can have another look.

Issue #53, opened 2020-05-28
import-graph dot file


jcommelin:
Currently, the dot file generated by `leanproject import-graph` contains lots of layout information. Of course this is useful for creating good-looking rendering. But it's annoying for post-processing the file using command-line filters like `sed` and `grep`. Could we have an option to generate a barebones minimalist .dot file?

jcommelin:
Alternatively, a `.tsort` file, that just creates a file in the format
```text
source1 target1
source2 target2
```
for every edge `(source, target)`.

PR #123, opened 2022-05-03
import-graph: write raw dot file to ".rawdot" without layout info

m4lvin:
fixes https://github.com/leanprover-community/mathlib-tools/issues/53

### DIFF
--- a/mathlibtools/import_graph.py
+++ b/mathlibtools/import_graph.py
@@ -16,6 +16,11 @@ class ImportGraph(nx.DiGraph):
         path = path or self.base_path/'import_graph.dot'
         nx.drawing.nx_pydot.to_pydot(self).write_dot(str(path))
 
+    def to_rawdot(self, path: Optional[Path] = None) -> None:
+        """Writes itself to a raw dot file (without layout)."""
+        path = path or self.base_path/'import_graph.rawdot'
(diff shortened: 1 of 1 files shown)


jcommelin:
This looks good to me.

jcommelin:
@m4lvin Do you know what's up with the failing check?

m4lvin:
I am afraid I do not know anything about the failing Pyinstaller check and now `The logs for this run have expired and are no longer available.` — is there a way to re-run the checks to investigate it?

jcommelin:
Merging master, or pushing an empty commit will certainly trigger CI to build again.

m4lvin:
I think the CI is failing due to some problem between pip and PyInstaller which is completely unrelated to the changes in the PR here. The other recent commit https://github.com/leanprover-community/mathlib-tools/commit/c24c0c5b5e10a0164263a2e20cba18dc964a86fe has exactly the same CI failure.

My knowledge of pip and python packages is too limited to debug this properly. I thought it might be related to the case-sensitive `pyinstaller` vs `PyInstaller` or https://github.com/pypa/pip/issues/6163 - in particular this says it should be fixed in pip 19.0.2 which was in 2019 and the github actions runner for windows seems to use pip 22.2.2 as per https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md

PR #126, opened 2022-10-13
feat: flag to exclude tactics in import-graph

semorrison:
This flag is helpful when determining what can be ported to mathlib4.

### DIFF
--- a/mathlibtools/import_graph.py
+++ b/mathlibtools/import_graph.py
@@ -70,6 +70,6 @@ class ImportGraph(nx.DiGraph):
     def exclude_tactics(self) -> 'ImportGraph':
         """Removes all files in src/tactic/ and src/meta/ from the graph."""
         H = self.subgraph([n for n,a in self.nodes(data=True) if
-          not a['label'].startswith('tactic') and not a['label'].startswith('meta')])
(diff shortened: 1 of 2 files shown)


mathlibtools/import_graph.py
```diff
@@ -66,3 +66,10 @@ def path(self, start: str, end: str) -> 'ImportGraph':
         H = self.subgraph(set(D.nodes).intersection(A.nodes))
         H.base_path = self.base_path
         return H
+
+    def exclude_tactics(self) -> 'ImportGraph':
+        """Removes all files in src/tactic/ and src/meta/ from the graph."""
+        H = self.subgraph([n for n,a in self.nodes(data=True) if
+          not a['label'].startswith('tactic') and not a['label'].startswith('meta')])
```
eric-wieser, line 73: Maybe
```suggestion
          not a['label'].startswith('tactic'.) and not a['label'].startswith('meta.')])
```
so as not to rule out third party names like `metamathematics`. Is `.` the separator or `/`?

semorrison: `.` is the separator: that's what displays when you view the graphs.

PatrickMassot:
Could you please rebase on master?

PR #128, opened 2022-10-19
Color node by port status

pechersky:
Use default wiki for yaml source

### DIFF
--- a/mathlibtools/file_status.py
+++ b/mathlibtools/file_status.py
@@ -31,11 +31,11 @@ class FileStatus:
 
   @classmethod
   def missing(cls) -> "FileStatus":
-    return cls({}, "orchid1")
+    return cls(set(), "orchid1")
 
   @classmethod
   def ready(cls) -> "FileStatus":
-    return cls({}, "turquoise1")
+    return cls(set(), "turquoise1")
 
   def matches(self, comment: str) -> bool:
     return all(substring.lower() in comment.lower() for substring in self.string_match)
--- a/mathlibtools/file_status.py
+++ b/mathlibtools/file_status.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Optional, Tuple
+from typing import Optional, Set
 
 @dataclass(frozen=True)
 class FileStatus:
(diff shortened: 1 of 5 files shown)


pechersky:
[algebra.order.ring.pdf](https://github.com/leanprover-community/mathlib-tools/files/9825044/algebra.order.ring.pdf)

mathlibtools/lib.py
```diff
@@ -1012,3 +1010,27 @@ def pull(self, remote: str='origin') -> None:
...
+        for filename, status in port_labels.items():
+            if filename not in self.import_graph.nodes:
+                continue
+            node = self.import_graph.nodes[filename]
+            if 'yes' in status.lower():
+                node["fillcolor"] = "green"
+                node["style"] = "filled"
+            elif 'no' in status.lower():
+                node["fillcolor"] = "red"
+                node["style"] = "filled"
```
semorrison, line 1036: Could we please have things that just say `No` not be colored at all?

But things that say `No` followed by one or more characters to be shown in yellow or orange?

Maybe we could use `PaleGreen` for yes, and `LightSalmon` for no-with-extra-info?

pechersky: [algebra.order.ring.pdf](https://github.com/leanprover-community/mathlib-tools/files/9825316/algebra.order.ring.pdf)

pechersky: I agree about no color on plain Nos. How about this?

semorrison: Actually, how about we are more specific about the `No`. Having a visual distinction between 'No: blocked' and 'No: WIP' is helpful. If it contains `WIP` or `PR`, than shade it something nice (maybe that same light blue?), and if it contains No and more characters, but not WIP or PR, then shade it orange.

pechersky:
[algebra.order.ring.pdf](https://github.com/leanprover-community/mathlib-tools/files/9825511/algebra.order.ring.pdf)
@semorrison  How about this? I followed your suggestions mostly. And refactored for easier modification.

mathlibtools/file_status.py
```diff
@@ -0,0 +1,51 @@
...
+    return cls({"yes"}, "green")
+
+  @classmethod
+  def pr(cls) -> "FileStatus":
+    return cls({"no", "pr"}, "lightskyblue")
+
+  @classmethod
+  def wip(cls) -> "FileStatus":
+    return cls({"no", "wip"}, "lightpink")
+
```
semorrison, line 27: Too many colours, I can't cope.

Let's have both of these be `lightskyblue`. They are both a form of "work is in progress", and if you want the details you just run `port_status.py`.

mathlibtools/file_status.py
```diff
@@ -0,0 +1,51 @@
...
+  @classmethod
+  def no(cls) -> "FileStatus":
+    return cls({"no"}, "orange")
+
+  @classmethod
+  def missing(cls) -> "FileStatus":
+    return cls(set(), "orchid1")
+
```
semorrison, line 35: And let's not display this. Most people don't want to know about files being missing from the port-status list. That's for bots to care about, not humans.

mathlibtools/file_status.py
```diff
@@ -0,0 +1,51 @@
...
+  @classmethod
+  def ready(cls) -> "FileStatus":
+    return cls(set(), "turquoise1")
+
+  def matches(self, comment: str) -> bool:
+    return all(substring.lower() in comment.lower() for substring in self.string_match)
```
semorrison, line 41: Oh, I don't like this. We care about the `yes` and `no` as initial strings, not anything afterwards.

semorrison:
It's not showing `logic.basic` as ready to port. Could you diagnose?

semorrison:
I like it!

PR #132, opened 2022-11-10
chore: fix port-status colour

joneugster:
As discussed on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/some.20useful.20automation/near/307016762).
`import_graph --port-status` currently ignores the [port status file](https://github.com/leanprover-community/mathlib/wiki/mathlib4-port-status/)  if all dependencies are ported. E.g. it shows `algebra.ne_zero` as "ready to port" instead of "PR exists".


joneugster:
Done in #133.

PR #133, opened 2022-11-11
chore(import-graph): don't overwrite 'PR' colour with 'ready' colour

semorrison:
The colours scheme in `import-graph` makes more sense if the "PR" status takes precedence of the "ready" colour.


semorrison:
Oh, I hadn't seen #132. This PR is subsumed by that one, although I don't immediately understand what else it is doing!

joneugster:
Please go ahead and merge this PR before #132! I didn't realise it would include any commits I made to my mathlib-tool fork automatically into the PR and used the same fork to do something different.

PR #134, opened 2022-11-12
feat: add transitive edges when removing tactics nodes from the import-graph

semorrison:
Previously if you used `--exclude-tactics` any imports that were only imported via transitive dependencies through tactics files would be silently dropped.


semorrison:
Also, changed `--exclude-tactics` so it leaves `tactic.basic` in place; otherwise the graphs are confusing to interpret.

PR #136, opened 2022-11-13
feat: import-graph --show-unused

semorrison:
`leanproject --to X --show-unused` will now display files which are not actually required by `X` in grey.

(In the sense that no declaration appears transitively, or via an ancestor, from any declaration in `X`.)

This will only work in mathlib, after https://github.com/leanprover-community/mathlib/pull/17507 is merged. If that is not cool, I can restructure things to include the relevant script in mathlib-tools, but this seems easier and sufficient (considering this tool is all end-of-life after the port?).

Example:

show-unused


digama0:
Hm, I would prefer showing the text in gray for this. IIRC we had some other meaning for red border? (File started in mathlib4 but not properly ported?)

semorrison:
Will switch to grey text. From what I recall the red outline was only used temporarily in something @pechersky was trying out.

pechersky:
Yes, red is file exists, code is in #137

semorrison:
Ah, okay. I am failing to work out how to do different color text on different nodes, however.

semorrison:
I propose to just shade unused modules light grey, and if you use `--port-status` and `--show-unused` simultaneously then `--port-status` will win if both want to color a node.

digama0:
is there an easy way to "lighten" the node / blend it with gray or white?

pechersky:
I believe it'll be easier to control the color if we merge the refactor

pechersky:
The color names that are used now are fixed X11 values, and one could do a `name -> rgba` conversion (via matplotlib or something) so that one could increase the alpha channel. Another option is to have two colors for nodes: https://stackoverflow.com/a/17654856. Yet another option is to use the [X11 colors](https://graphviz.org/doc/info/colors.html) that end in `3` or `4`, and to "lighten", change that to a `1`.

PR #137, opened 2022-11-13
refactor FileStatus as class, add `--mathlib4`

pechersky:
To further support mathlib porting, the `FileStatus` tracker is refactored.
This will allow for better parsing long-term, easier node annotation (like href linking).

For QOL, one can pass a `mathlib4` path to color node outlines as red for existing files.


eric-wieser:
Can you fix the conflicts (and get the new CI config in the process)?

eric-wieser:
Does this affect the format of the yaml data? If so, can you paste the new format here?

pechersky:
This PR on its own does not affect the format yet. The code as it is right now supports parsing the current format. Once we merge this, then we can generate the new format and have a "PR" for switching the wiki format.

mathlibtools/file_status.py
```diff
@@ -1,53 +1,66 @@
...
+        return {k: v for k, v in asdict(self).items() if v is not None}
+
+    def color(self) -> Optional[str]:
+        if self.ported:
+            return "green"
+        if self.mathlib4_pr:
+            return "lightskyblue"
+        if self.comments and "ready" in self.comments:
+            return "turquoise1"
+        return None
```
eric-wieser, line 44: In my opinion this belongs in the `port_status()` in the other file; colors are a property of the graphviz output, not of the data representation

mathlibtools/file_status.py
```diff
@@ -1,53 +1,66 @@
...
+
+
+@dataclass
+class PortStatus:
```
eric-wieser, line 48: Can you add a single API function that is do everyhing at once? For CI, I want a single function to call that is "load the portstatus from the wiki" so that I can process it. If there are multiple functions I'm more likely to be broken by future updates.

pechersky: `deserialize_old` will be that API for now. I made it have the input yaml as optional, so it does what you expect by default

semorrison:
bors merge

semorrison:
@pechersky, can you ping Patrick if we want an immediate release?

eric-wieser:
We need #146 before we can commit to this new API, otherwise there's data in the yaml we can't parse

PR #144, opened 2022-11-16
GEXF export with FileStatus

pechersky:
Convert the FileStatus to a str (using `repr`),
and absent the FileStatus, an empty string

### DIFF
--- a/mathlibtools/file_status.py
+++ b/mathlibtools/file_status.py
@@ -1,6 +1,6 @@
+import re
 from dataclasses import asdict, dataclass
 from typing import Any, Dict, Optional, Union
-import re
 
 import requests
 import yaml
--- a/mathlibtools/file_status.py
+++ b/mathlibtools/file_status.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from typing import Optional, Set
+from typing import List, Optional, Set, Union
 
 @dataclass(frozen=True)
 class FileStatus:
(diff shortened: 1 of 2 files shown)


PatrickMassot:
I'm sorry I had very little time and now this PR has serious conflict. Yakov, could you try to update your PR please?

pechersky:
Sorry for the wait since your message, Patrick, I've had a pretty bad fever this week.

PatrickMassot:
No problem, thanks for updating!

mathlibtools/file_status.py
```diff
@@ -50,6 +50,9 @@ def pr_link(self) -> Optional[str]:
             return None
         return f"https://github.com/leanprover-community/mathlib4/pull/{self.mathlib4_pr}"
 
+    def to_gexf(self) -> str:
+        return repr(self)
```
eric-wieser, line 54: What's your plan for parsing this in the consumer? It might make more sense to unpack it into a series of separate attributes, or maybe serializing to json in some way.

pechersky: I don't know what downstream gexf uses are -- Patrick brought up to me that gexf export was completely blocked by file statuses. The export needs a very basic type, str or numeric, doesn't even support lists. So I think one solution would have been to just strip statuses completely in the gexf export.

If an end user has a gexf application/use, I hope they file an issue on making the output easier to parse.

eric-wieser: Ah that makes sense, thanks!

PR #145, opened 2022-11-16
Fix typing on `path` to rid of mypy lint errors

pechersky:
Otherwise, mypy can't infer that `to: str`.


PatrickMassot:
This was actually merged somewhere else.

PR #146, opened 2022-11-19
Parse comments in FileStatus

eric-wieser:


semorrison:
bors merge

eric-wieser:
There is no bors here

semorrison:
Ah, good, because I noticed a problem. Doesn't `port_status()` in `lib.py` clobber the comments field?

PR #151, opened 2022-11-22
fix: port-progress reports remaining properly

pechersky:


mathlibtools/leanproject.py
```diff
@@ -373,6 +373,8 @@ def port_progress(to: Optional[str]) -> None:
     proportion_lines = round(nb_ported_lines/nb_lines*100, 1)
     longest_unported_path = graph.longest_path_length()
     progress_path = round(100 - longest_unported_path / mathlib3_longest_path * 100, 1)
+    if not to:
```
eric-wieser, line 376:
```suggestion
    if to is None:
```
feels more explicit

eric-wieser (approved):

semorrison:
Looks good, running locally.

